PR #22, opened 2020-09-24
Add schdule cache feature.

mizhenkun:
Add schedule cache

### DIFF
--- a/test/sched_sample.c
+++ b/test/sched_sample.c
@@ -47,6 +47,82 @@ struct sample_sched_ctx {
 	struct sample_sched_info sched_info[0];
 };
 
+struct cache {
+    __u32 *buff;
+    __u32 depth;
+    __u32 head;
+    __u32 tail;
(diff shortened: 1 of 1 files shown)


hzhuang1 (commented):
It's better to split it into two small patches. One is to fix in test_hisi_sec.c, and the other is for schedule cache.

If you can add more comments in commit for schedule cache, it would be better. It could help other developers/users to understand how it works.

mizhenkun:
> It's better to split it into two small patches. One is to fix in test_hisi_sec.c, and the other is for schedule cache.
> 
> If you can add more comments in commit for schedule cache, it would be better. It could help other developers/users to understand how it works.

I delete the change of test_hisi_sec.c, and add some commit info.

wangzhou:
better to write subject line:

test: sched: xxx

mizhenkun:
> better to write subject line:
> 
> test: sched: xxx

Fixed

PR #24, opened 2020-09-24
hpre: fix coding style

barry19901226:
fix coding style

### DIFF
--- a/test/hisi_sec_test/test_hisi_sec.c
+++ b/test/hisi_sec_test/test_hisi_sec.c
@@ -60,14 +60,13 @@ typedef struct wd_thread_res {
 	unsigned long long send_num;
 	unsigned long long recv_num;
 	struct timeval start_tval;
+	unsigned long long sum_perf;
 } thread_data_d;
 
 //static pthread_cond_t cond = PTHREAD_COND_INITIALIZER;
 static pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
 static pthread_t system_test_thrds[NUM_THREADS];
 static thread_data_t thr_data[NUM_THREADS];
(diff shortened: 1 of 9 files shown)


test/hisi_sec_test/test_hisi_sec.c
```diff
@@ -1744,8 +1745,12 @@ int main(int argc, char *argv[])
 	if (!strcmp(argv[3], "-optype")) {
 		g_direction = strtoul((char*)argv[4], NULL, 10);
 		if (algtype_class == DIGEST_CLASS) {
```
hzhuang1, line 1747: Why does the mode number in the range of 0, 1 and 3? Do we need mode number 2?

test/hisi_sec_test/test_hisi_sec.c
```diff
@@ -1744,8 +1745,12 @@ int main(int argc, char *argv[])
...
-			//0 is normal mode, 1 is HMAC mode.
+			//0 is normal mode, 1 is HMAC mode, 3 is long hash mode.
 			alg_op_type = g_direction;
+			if (g_direction == 3) {
+				alg_op_type = 0;
+				g_ivlen = 1;
+			}
```
hzhuang1, line 1753: It's not related to perf. Could you help moving it to another patch?

wangzhou (commented):
fine to me

test/hisi_sec_test/test_hisi_sec.c
wangzhou: we should use getopt_long to do this

wangzhou: maybe 2 will be used for other mode ：）

PR #25, opened 2020-09-25
docs: append INSTALL doc

hzhuang1:
Add instructions on building UADK and running test cases in INSTALL document.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -1,0 +1,77 @@
+Deploying UADK
+
+UADK is a framekwork for user application to access hardware, and it depends
+UACCE in kernel.
+
+
+1. Kernel Requirement
+
+User needs to make sure that UACCE is already supported in Linux kernel. The
+kernel version should be at least v5.9 with SVA (Shared Virtual Addressing)
+enabled.
+
+
+2. Kernel Configuration
+
+UACCE could be built as module or built-in.
+
+Here's an example to enable UACCE with hardware accelerator in Hisilicon
+platform.
+
+    CONFIG_IOMMU_SVA_LIB=y
+    CONFIG_ARM_SMMU=y
+    CONFIG_ARM_SMMU_V3=y
+    CONFIG_ARM_SMMU_V3_SVA=y
+    CONFIG_PCI_PASID=y
+    CONFIG_UACCE=y
+    CONFIG_CRYPTO_DEV_HISI_SEC2=y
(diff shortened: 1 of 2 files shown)


INSTALL
```diff
@@ -0,0 +1,69 @@
+Deploying UADK
+
+UADK is a framekwork for user application to access hardware, and it depends
+UACCE in kernel.
+
+
+1. Kernel Requirement
+
+User needs to make sure that UACCE is already supported in Linux kernel. The
+kernel version should be at least v5.9.
```
wangzhou, line 10: In fact, we need a SVA kernel, so v5.9 is not enough :)

INSTALL
```diff
@@ -0,0 +1,69 @@
...
+UACCE could be built as module or built-in.
+
+Here's an example to enable UACCE with hardware accelerator in Hisilicon
+platform.
+
+    CONFIG_UACCE=y
+    CONFIG_CRYPTO_DEV_HISI_SEC2=y
+    CONFIG_CRYPTO_DEV_HISI_QM=y
+    CONFIG_CRYPTO_DEV_HISI_ZIP=y
+    CONFIG_CRYPTO_DEV_HISI_HPRE=y
```
wangzhou, line 24: let's also mention SVA, SMMU, PCI_PASID related configures

INSTALL
```diff
@@ -0,0 +1,69 @@
...
+    Asynchronous mode
+    $ sudo ./test/hisi_zip_test/test_sva_perf -b 8192000 -l 1000 -v -m 1
+
+Test on sec hardware accelerator
+
+    Synchronous mode
+    $ sudo ./test/hisi_sec_test/test_hisi_sec -cipher 0 -optype 0 -pklen 16 -keylen 16 -times 1 -sync -multi 1
+    Asynchronous mode
+    $ sudo ./test/hisi_sec_test/test_hisi_sec -cipher 0 -optype 0 -pklen 16 -keylen 16 -times 1 -async -multi 1
+
```
wangzhou, line 63: let's also mention digest test case, in fact, just replace -cipher above to -digest

hzhuang1: fixed

INSTALL
hzhuang1: fixed

INSTALL
hzhuang1: fixed

PR #33, opened 2020-09-28
wd/sec: modify cipher and digest driver code

Liulongfang:
1.fixed some driver coding styles
2.fixed some test code
3.sync the bug fix code to the uadk from CI


Liulongfang:
Merging is blocked

PR #34, opened 2020-09-28
wd/sec: modify cipher and digest driver code

Liulongfang:
1.fixed some driver coding styles
2.fixed some test code
3.sync the bug fix code to the uadk from CI


hzhuang1 (commented):
It's better to format two patches. One is to fix code, and the other is to fix coding style.

hzhuang1:
It's better not using words as "sync from CI". You need to explain what you are doing in the commit. Other developers can't access your internal code base, it makes nonsense to them with these words. They need to understand why this fix is necessary in UADK.

PR #49, opened 2020-09-30
wd/sec: modify AEAD async driver code

Liulongfang:
Added the AEAD aiv addr for the async mode

### DIFF
--- a/drv/hisi_sec.c
+++ b/drv/hisi_sec.c
@@ -1036,6 +1036,7 @@ static void parse_aead_bd2(struct hisi_sec_sqe *sqe,
 		recv_msg->result = WD_SUCCESS;
 	}
 
+	recv_msg->aiv = (__u8 *)(uintptr_t)sqe->type2.a_ivin_addr;
 	recv_msg->tag = sqe->type2.tag;
 
 #ifdef DEBUG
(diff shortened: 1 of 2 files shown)


wd_aead.c
```diff
@@ -546,6 +546,12 @@ int wd_do_aead_async(handle_t h_sess, struct wd_aead_req *req)
 	}
 
 	fill_request_msg(msg, req, sess);
+	msg->aiv = malloc(req->iv_bytes);
+	if (req->iv_bytes == 0 || !msg->aiv) {
```
wangzhou, line 550: iv_bytes == 0 should be checked before.

drv/hisi_sec.c
```diff
@@ -1036,6 +1036,7 @@ static void parse_aead_bd2(struct hisi_sec_sqe *sqe,
 		recv_msg->result = WD_SUCCESS;
 	}
 
+	recv_msg->aiv = (__u8 *)(uintptr_t)sqe->type2.a_ivin_addr;
```
wangzhou, line 1039: I do not see any usage in aead algorithm layer about aiv. why do we need this here?

Liulongfang: aiv come from iv, and Hardware need aiv

PR #51, opened 2020-10-12
wd_comp: optimize on wd_do_comp_sync2()

hzhuang1:
1. Merge duplicated code. Most code could be shared between compression
and decompression.
2. Update avail_out for each transition.
3. The length of each line should be limited in 80 characters.


haofang111 (commented):
good. +1

PR #59, opened 2020-10-15
wd: fixes some bugs for SEC

Liulongfang:
1. add poll parameter check 
2. update SEC algorithm specifications 
3. add ctx mode check for SEC 
4.modify API uninit interface


Liulongfang:
need mofiy

PR #60, opened 2020-10-15
wd: fixes some bugs for SEC

Liulongfang:
1.add poll parameter check
2.update SEC algorithm specifications
3.add ctx mode check for SEC
4.modify API uninit interface

### DIFF
--- a/include/wd_alg_common.h
+++ b/include/wd_alg_common.h
@@ -2,6 +2,7 @@
 #define WD_ALG_COMMON_H
 
 #include <asm/types.h>
+#include <pthread.h>
 #include "wd.h"
 
 #ifndef WD_ERR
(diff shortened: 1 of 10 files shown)


include/wd_alg_common.h
```diff
@@ -2,6 +2,7 @@
 #define WD_ALG_COMMON_H
 
 #include <asm/types.h>
+#include <pthread.h>
```
hzhuang1, line 5: Why do you need to add pthread head file at here? I can't find the dependency.

hzhuang1 (commented):
It's better to write comments. It's not good to leave comments as blank in one commit.

include/wd_alg_common.h
Liulongfang: This demand is required by Wangzhou

PR #61, opened 2020-10-15
test: sec: Delete libnuma path when static compile

mizhenkun:


Makefile.am
```diff
@@ -32,7 +32,7 @@ libwd_crypto_la_SOURCES=wd_cipher.c wd_cipher.h wd_cipher_drv.h \
 			wd_dh.c wd_dh.h wd_dh_drv.h \
 			wd_digest.c wd_digest.h wd_digest_drv.h \
 			wd_util.c wd_util.h
-libwd_crypto_la_LIBADD= $(libwd_la_OBJECTS) -ldl
+libwd_crypto_la_LIBADD= $(libwd_la_OBJECTS) -ldl -lnuma
```
hzhuang1: I remember that it will cause build issue if "-lnuma" is added in "libwd_crypto_la_LIBADD". So you applied a workaround on linking numa library only in test app. Have you fixed the issue?

hzhuang1:
You need to rebase the patches since there's a new merge commit generated in your patch set.

Makefile.am
mizhenkun: No, add -lnuma at this location can't fixed the issue.  I fixed the issue as follows:
image

mizhenkun:
> You need to rebase the patches since there's a new merge commit generated in your patch set.

Done

hzhuang1:
I'm fine on your latest patch. But the title mentioned "-Bstatic" that not existed in your patch. So the title doesn't match the patch content. Could you help to fix this, too?

PR #66, opened 2020-10-26
test: sec: Add the ctx num para in sec test app

mizhenkun:
For support the multiple ctxs test case, the test app
support the ctx option for user.We can input the value
to control app alloc multi hareware queue resource and
procsess the data.

Signed-ff by: Zhenkun Mi <mi_zhenkun@163.com>

### DIFF
--- a/test/hisi_sec_test/test_hisi_sec.c
+++ b/test/hisi_sec_test/test_hisi_sec.c
@@ -29,6 +29,7 @@ static struct wd_sched *g_sched;
 
 static long long int g_times;
 static unsigned int g_thread_num;
+static unsigned int ctx_num = 0;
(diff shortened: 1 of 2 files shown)


test/hisi_sec_test/test_hisi_sec.c
```diff
@@ -289,36 +291,32 @@ static int init_sigle_ctx_config(int type, int mode)
 {
```
hzhuang1, line 291: Single in init_single_ctx_config() means the single context. The original meaning is changed after multiple contexts are supported. You need to change the function name, too.

mizhenkun: Fixed.

PR #73, opened 2020-11-04
wd: mempool: Add the mempool block pool funtion

mizhenkun:
To solve the page fault in the kernel, uadk will support the
memory pin feature. The user app need a mechanism to manage
the memory, we name it wd_mempool, this patch finish one part
of the mempool, blk_pool.

Signed-ff by: Zhenkun Mi <mi_zhenkun@163.com>

### DIFF
--- a/wd_mempool.c
+++ b/wd_mempool.c
@@ -1,0 +1,105 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+#include <ctype.h>
+#include <dirent.h>
+#include <errno.h>
+#include <poll.h>
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/ioctl.h>
+#include <sys/mman.h>
+#include "wd.h"
+
+struct blkpool {
+	void **blk_elem; /* all the block unit addrs saved in blk_elem */
+	size_t depth;    /* the block pool deph, stack depth */
+	size_t top;      /* the stack top pos for blk_elem */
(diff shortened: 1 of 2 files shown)


wd_mempool.c
```diff
@@ -0,0 +1,105 @@
...
+
+void *wd_blockpool_alloc(handle_t blockpool)
+{
+	struct blkpool *bp = (struct blkpool*)blockpool;
+	
+	if (!bp)
+		return NULL;
+
+	if (bp->top > 0) {
+		bp->top--;
```
hzhuang1, line 34: top is either increased by 1 or decreased by 1. The behavior of top is much like an index. So we could apply "int" type on the "top" field.

Same on "depth" field.

wd_mempool.c
```diff
@@ -0,0 +1,105 @@
...
+
+void wd_blockpool_free(handle_t blockpool, void *addr)
+{
+	struct blkpool *bp = (struct blkpool*)blockpool;
+	
+	if (!bp || !addr)
+		return;
+
+	if (bp->top < bp->depth) {
+		bp->blk_elem[bp->top] = addr;
```
hzhuang1, line 49: If there're four blocks, all these four blocks (block 0-3) are allocated by wd_blockpool_alloc(). Now block 3 is freed by wd_blockpool_free().
At this time, bp->top is 0. The address of block 3 will be filled into block 0 since "top" field is still 0.

And why should we update "blk_elem" field at here? It seems that nobody will clean it.

PR #77, opened 2020-11-13
Fix misc

hzhuang1:


wangzhou (approved):
So we should also fix uninit_config? Let's first merge this patch, and later fix uninit_config :)

PR #85, opened 2020-11-23
Lib version

hzhuang1:


conf.sh
```diff
@@ -14,14 +14,25 @@ if [ $1 ]; then
...
 	fi
 fi
 
+if [ $2 ]; then
+	if [ $2 = "--with-uadk_v1" ]; then
+		echo "enable UADK v1!"
+		UADK_VERSION="--with-uadk_v1"
+	fi
+fi
+
```
wangzhou, line 31: confused here. Could we add some comments about the v1/v2, static/shared lib conf.sh command?

configure.ac
```diff
@@ -94,9 +89,19 @@ AC_FUNC_MALLOC
...
+			 test/hisi_sec_test/Makefile
+			 test/hisi_zip_test/Makefile])
+)
+
+#AC_CONFIG_FILES([Makefile
+#         test/Makefile
+#	 test/hisi_hpre_test/Makefile
+#	 test/hisi_sec_test/Makefile
+#         test/hisi_zip_test/Makefile
+#	 v1/Makefile])
```
wangzhou, line 109: can we remove above comments?

configure.ac
```diff
@@ -94,9 +89,19 @@ AC_FUNC_MALLOC
...
-		 test/hisi_sec_test/Makefile
-                 test/hisi_zip_test/Makefile])
+AC_CONFIG_FILES([Makefile])
+AS_IF([test -n "$with_uadk_v1"],
+	AC_CONFIG_FILES([v1/Makefile]),
+	AC_CONFIG_FILES([test/Makefile
+			 test/hisi_hpre_test/Makefile
+			 test/hisi_sec_test/Makefile
+			 test/hisi_zip_test/Makefile])
+)
```
wangzhou, line 102: I am not sure if we still need this?

v1/test/Makefile.am
```diff
@@ -1,18 +1,18 @@
...
-test_hisi_zlib_LDADD=../.libs/libwd.a
-wd_zip_test_LDADD=../.libs/libwd.a lib/libz.a
+test_hisi_zip_LDADD=../.libs/libwd.la
+test_hisi_zlib_LDADD=../.libs/libwd.la
+wd_zip_test_LDADD=../.libs/libwd.la -lz
 
 test_dummy_SOURCES=test_dummy.c
 
-test_dummy_LDADD=../.libs/libwd.a
+test_dummy_LDADD=../.libs/libwd.la
```
wangzhou: we can remove dummy test at all

conf.sh
hzhuang1: OK. I'll update it with comments.

configure.ac
hzhuang1: No, we can't remove them. Without them, Makefile can't be generated successfully.

configure.ac
hzhuang1: Yes, we need them.

v1/test/Makefile.am
hzhuang1: OK. I'll remove those dummy tests.

PR #100, opened 2020-12-03
wd/sec: add new algorithms for Kunpeng930

Liulongfang:
Add new BD type3 algorithms for SEC driver
and solved the driver compatibility issues.

### DIFF
--- a/drv/hisi_qm_udrv.c
+++ b/drv/hisi_qm_udrv.c
@@ -75,6 +75,10 @@ static struct hisi_qm_type qm_type[] = {
 		.qm_name	= "hisi_qm_v2",
 		.qm_db_offs	= QM_V2_DOORBELL_OFFSET,
 		.hacc_db	= hacc_db_v2,
+	}, {
+		.qm_name	= "hisi_qm_v3",
+		.qm_db_offs	= QM_V2_DOORBELL_OFFSET,
+		.hacc_db	= hacc_db_v2,
 	},
 };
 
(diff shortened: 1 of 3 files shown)


hzhuang1:
The v3 is just copied from v2. So there's no difference between v2 and v3. Why do we need this patch?

hzhuang1:
Sorry. I just clicked the wrong button. So re-open it.

Liulongfang:
> The v3 is just copied from v2. So there's no difference between v2 and v3. Why do we need this patch?

no, the driver of v3 is completely different from v2.

drv/hisi_qm_udrv.c
```diff
@@ -75,6 +75,10 @@ static struct hisi_qm_type qm_type[] = {
 		.qm_name	= "hisi_qm_v2",
 		.qm_db_offs	= QM_V2_DOORBELL_OFFSET,
 		.hacc_db	= hacc_db_v2,
+	}, {
+		.qm_name	= "hisi_qm_v3",
+		.qm_db_offs	= QM_V2_DOORBELL_OFFSET,
+		.hacc_db	= hacc_db_v2,
```
wangzhou, line 81: I think this is wrong, I should think hardware will not change in futurn,  so v2 can run for futurn v3, v4, v5.

Liulongfang: I need the version handle to compatible interface：

PR #101, opened 2020-12-03
Fix test

hzhuang1:


test/sanity_test.sh
```diff
@@ -6,6 +6,12 @@
 # steps in INSTALL doc.
 #
 # Please feel free to add more basic tests.
+#
+# If user doesn't install the binaries, libraries and head files into the
+# configed path, user must specify path in this way.
```
fboudra, line 11: typo: configed -> configured

test/sanity_test.sh
```diff
@@ -6,6 +6,12 @@
...
+#
+# $sudo LD_LIBRARY_PATH={library path} PATH={bin path}	\
+# C_INCLUDE_PATH={head path} sanity_test.sh
```
fboudra, line 14: not sure we need C_INCLUDE_PATH because we need only runtime variables (LD_LIBRARY_PATH and PATH)

test/sanity_test.sh
```diff
@@ -14,6 +20,31 @@ zip_result=-1
...
 hpre_result=-1
 
+TEST_FILE=test_uadk_lib.c
+
+# failed: return 1; success: return 0
+check_uadk_lib()
+{
+	# wd_get_acce_list() is available in v2. Use this function to test
+	# whether libwd.so is v2.
+	echo "#include <stdio.h>" > $TEST_FILE
```
fboudra, line 30: instead of using several echos, use a heredocument:

cat << EOF > ${TEST_FILE}
#include ...
EOF

fboudra (commented):
LGTM on the principle. Just 3 comments on improvement that you can make. Thanks!

test/sanity_test.sh
hzhuang1: Fixed.

test/sanity_test.sh
hzhuang1: Yes, we need it since the test file need to include <wd.h>. This "wd.h" is installed from uadk. Without this head file, we'll meet compilation issue.

test/sanity_test.sh
hzhuang1: Fixed.

PR #147, opened 2020-12-31
drv: qm: Add support of user space (v)SVA task resetting

yekai123123:
While app run on ACC PF and VF in user space including
VM (vSVA) user space, they can be notified as RAS resetting.
Afer that, app can request queue end release the old queue, and
do the task on the new queue again.

### DIFF
--- a/include/hisi_qm_udrv.h
+++ b/include/hisi_qm_udrv.h
@@ -42,6 +42,8 @@ struct hisi_qm_queue_info {
 	void *db_base;
 	int (*db)(struct hisi_qm_queue_info *q, __u8 cmd,
 		  __u16 index, __u8 priority);
+	void *ds_tx_base;
+	void *ds_rx_base;
 	__u16 sq_tail_index;
 	__u16 sq_head_index;
 	__u16 cq_head_index;
@@ -51,6 +53,7 @@ struct hisi_qm_queue_info {
 	__u16 hw_type;
 	bool cqc_phase;
 	pthread_spinlock_t lock;
+	unsigned long region_size[UACCE_QFRT_MAX];
 };
 
 struct hisi_qp {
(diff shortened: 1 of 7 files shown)


include/wd.h
```diff
@@ -304,4 +304,13 @@ extern void wd_free_list_accels(struct uacce_dev_list *list);
  */
 extern int wd_ctx_set_io_cmd(handle_t h_ctx, unsigned long cmd, void *arg);
 
+/**
+ * wd_get_dev_region_size() - Get region offset size
```
wangzhou, line 308: Get region size

wd.c
```diff
@@ -366,6 +366,14 @@ void wd_drv_unmap_qfr(handle_t h_ctx, enum uacce_qfrt qfrt)
 		munmap(ctx->qfrs_base[qfrt], ctx->qfrs_offs[qfrt]);
 }
 
+unsigned long wd_get_dev_region_size(handle_t h_ctx, enum uacce_qfrt qfrt)
```
wangzhou, line 369: I think better to be wd_ctx_get_region_size

include/hisi_qm_udrv.h
```diff
@@ -51,6 +53,7 @@ struct hisi_qm_queue_info {
 	__u16 hw_type;
 	bool cqc_phase;
 	pthread_spinlock_t lock;
+	unsigned long qfrs_offset[UACCE_QFRT_MAX];
```
wangzhou, line 56: qfrs_size

wangzhou, line 56: BTW, why we need qfrs_offset, we can set ds_tx_base and ds_rx_base directly.

And what is ds_tx_base/ds_rx_base, please get a meanful name, and add comments here

yekai123123:
abandon it

PR #153, opened 2021-01-04
drv: qm: Add support of user space (v)SVA task resetting

yekai123123:
While app run on ACC PF and VF in user space including
VM (vSVA) user space, they can be notified as RAS resetting.
Afer that, app can request queue end release the old queue, and
do the task on the new queue again.

### DIFF
--- a/wd.c
+++ b/wd.c
@@ -366,6 +366,14 @@ void wd_drv_unmap_qfr(handle_t h_ctx, enum uacce_qfrt qfrt)
 		munmap(ctx->qfrs_base[qfrt], ctx->qfrs_offs[qfrt]);
 }
 
+unsigned long wd_ctx_get_region_size(handle_t h_ctx, enum uacce_qfrt qfrt)
+{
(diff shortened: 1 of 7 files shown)


wd.c
```diff
@@ -366,6 +366,14 @@ void wd_drv_unmap_qfr(handle_t h_ctx, enum uacce_qfrt qfrt)
...
 }
 
+unsigned long wd_ctx_get_region_size(handle_t h_ctx, enum uacce_qfrt qfrt)
+{
+	struct wd_ctx_h *ctx = (struct wd_ctx_h *)h_ctx;
+	if (!ctx || qfrt >= UACCE_QFRT_MAX)
+			return 0;
+	return ctx->qfrs_offs[qfrt];
+}
+
```
hzhuang1, line 376: A new API is added in libwd. How do you use it? Could it be used in your test case?

yekai123123: Qm user driver needs region size as device resetting,  but driver layer can not find ctx, so add a new API to get region size based h_ctx.

PR #196, opened 2021-01-28
test: sec: clean thread data in sec_sva_test()

hzhuang1:
Two kinds of thread data are used in sec_sva_test(),
sva_sync_create_threads() and sva_async_create_threads().

Fix to only use the thread data defined in sec_sva_test(). Avoid to use
the global thread data in both sva_sync_create_threads() and
sync_async_create_threads().

### DIFF
--- a/test/hisi_sec_test/test_hisi_sec.c
+++ b/test/hisi_sec_test/test_hisi_sec.c
@@ -3146,8 +3146,8 @@ static void *sva_poll_func(void *arg)
 	return NULL;
 }
 
-static int sva_async_create_threads(int thread_num, struct wd_cipher_req *reqs,
-	struct wd_cipher_sess_setup *setups, thread_data_t *tds)
+static int sva_async_create_threads(int thread_num,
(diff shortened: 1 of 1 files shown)


test/hisi_sec_test/test_hisi_sec.c
```diff
@@ -3384,12 +3375,16 @@ static int sec_sva_test(void)
 		}
 		req[i].cb = async_cb;
 		req[i].cb_param = &datas[i];
+
+		datas[i].tid = i;
+		datas[i].req = &req[i];
+		datas[i].setup = &setup[i];
```
wangzhou, line 3381: miss bd_pool?

hzhuang1: bd_pool is already set by init_bd_pool() function.

thr_data[] are global variable array, and datas[] are internal variable array. A lot of initializations are done in datas array. If we switch to the internal array, we don't need to copy bd_pool to thr_data array any more.

PR #199, opened 2021-01-30
errors && clean code

yumeng0117:
1. UADK/HPRE/RSA - modyfy rsa error found by ut;
2. UADK - clean code.


hzhuang1:
Please help to update the title.
The first one could be "wd: rsa: xxx".
The second one could be "hisi: hpre: xxx".
And please help to fill the comments. Don't leave it as blank.

PR #203, opened 2021-02-02
Sync code clean for wd and qm

lin755:
Signed-off-by: wenkai lin <linwenkai755@qq.com>

### DIFF
--- a/drv/hisi_hpre.c
+++ b/drv/hisi_hpre.c
@@ -1188,9 +1188,9 @@ static int ecc_prepare_in(struct wd_ecc_msg *msg,
 
 	switch (msg->req.op_type) {
 	case HPRE_SM2_ENC: /* fall through */
-	case HPRE_SM2_DEC: /* fall through */
+	case HPRE_SM2_DEC:
 		/* driver to identify sm2 algorithm when async receive */
-		hw_msg->sm2_mlen = msg->req.op_type;
+		hw_msg->sm2_mlen = msg->req.op_type; /* fall through */
 	case WD_SM2_KG: /* fall through */
 	case WD_ECXDH_GEN_KEY:
 		ret = ecc_prepare_dh_gen_in(msg, hw_msg, data);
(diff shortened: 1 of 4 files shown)


hzhuang1:
Don't use dash in the title. Please help to use colon instead.

Don't leave the comment area as blank.

lin755:
this patch for:
1.make fall through effective
2.some clean for size_t
3.tmp use directly in next ,add intercept

hzhuang1:
For the title, we always write it as the format "{category}: {operation}". For example, "wd: clean for size_t". If it's related to rsa algorithm, we could write "wd: rsa: clean for size_t" or "wd/rsa: clean for size_t".

It's good that you fill the comment in the pull request. But I really hope you could move it to the comment area in the commit. It would help others to check comments in git tree. If the comments are only in the pull request, others couldn't find it from the git tree.

To your patch, it's better to split it since it covers both wd common layer and hisi vendor driver. If a change is related to any interface, we have to make the changes on both wd layer and vendor driver into one commit. Otherwise, we'll suffer build error. If a change is NOT related to any interface, we could split the patch. The rule is to only change a small part at a time. It would be easier to review in the community.

Thanks.

lin755:
> For the title, we always write it as the format "{category}: {operation}". For example, "wd: clean for size_t". If it's related to rsa algorithm, we could write "wd: rsa: clean for size_t" or "wd/rsa: clean for size_t".
> 
> It's good that you fill the comment in the pull request. But I really hope you could move it to the comment area in the commit. It would help others to check comments in git tree. If the comments are only in the pull request, others couldn't find it from the git tree.
> 
> To your patch, it's better to split it since it covers both wd common layer and hisi vendor driver. If a change is related to any interface, we have to make the changes on both wd layer and vendor driver into one commit. Otherwise, we'll suffer build error. If a change is NOT related to any interface, we could split the patch. The rule is to only change a small part at a time. It would be easier to review in the community.
> 
> Thanks.

Thanks for all your advices, I will send another pr and close this later.

hzhuang1:
Close since the patches are migrated into another pull request #205.

PR #206, opened 2021-02-04
Misc

hzhuang1:


Liulongfang:
Run the test case and the result is correct to ensure that the modification is OK.

hzhuang1:
I've appended sglnum parameter in sanity. So it could be always verified.

Liulongfang (approved):

PR #211, opened 2021-02-09
uadk/sec: fixup deadlock proble in qm send process and some spelling …

yekai123123:
…error

1. fixup deadlock problem during multi threads flr and control reset in qm send.
2. fixup spelling error on comments and function name.


hzhuang1:
Actually you didn't fix on any deadlock issue. You're just fixing the spelling error. Let's rewrite the subject. For example, "wd: aead: fix typo issue".

And split the patch into smaller ones. Some for vendor drivers, and others for algorithm layer.

PR #224, opened 2021-03-03
wd: make UADK v2 compatible with v1

hzhuang1:
Make UADK library include both interfaces from v1 and v2.


haofang111:
ok

PR #236, opened 2021-03-23
wd: add wd_ctx_wait to sync mode

lin755:
Since mainline kernel not upload poll now, we set timeout to 0 for temp hack

Signed-off-by: wenkai lin <linwenkai755@qq.com>

### DIFF
--- a/wd_aead.c
+++ b/wd_aead.c
@@ -19,6 +19,9 @@
 #define DES_WEAK_KEY_NUM	4
 #define MAX_RETRY_COUNTS	200000000
 
+#define POLL_SIZE		70000
+#define POLL_TIME		0
+
 static __u64 des_weak_key[DES_WEAK_KEY_NUM] = {
 	0x0101010101010101, 0xFEFEFEFEFEFEFEFE,
 	0xE0E0E0E0F1F1F1F1, 0x1F1F1F1F0E0E0E0E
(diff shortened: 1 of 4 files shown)


wd_aead.c
```diff
@@ -520,6 +523,14 @@ int wd_do_aead_sync(handle_t h_sess, struct wd_aead_req *req)
...
 	}
 
+	if (req->in_bytes >= POLL_SIZE) {
+		ret = wd_ctx_wait(ctx->ctx, POLL_TIME);
+		if (ret < 0) {
+			WD_ERR("wd ctx wait err(%d)!\n", ret);
+			goto err_out;
+		}
+	}
+
```
hzhuang1, line 533: Do you mean that adopt polling if source buffer is larger than POLL_SIZE limitation? Otherwise, the source buffer is small, and use busy-loop wait to receive output from hardware. Is it right?

lin755: If req in_bytes is bigger，it means need more time for hardware to do one task, I pick this POLL_SIZE from 920 test data, we use poll wait instead of busy-loop to reduce CPU usage.But this function need kernel support poll, do you think it is better to support it after mainline upload poll？

hzhuang1: While your patch is applied, will it improve performance?
From current data, synchronous mode gain much more performance on asynchronous mode. If you could observe the same result, I suggest hold it on until the performance on asynchronous mode is improved.

lin755: This patch will not improve performance,its main purpose was  to reduce cpu usage for sync mode.but I can still try to use it for async mode, i will close this request temporarily

PR #250, opened 2021-04-17
Some patchs of comp

youngersh:


wangzhou (commented):
-l currently is using compact case, please do not use it

youngersh:
> -l currently is using compact case, please do not use it

I will fix this in next version. Thx!

PR #257, opened 2021-04-26
uadk: wd - add libwd.a version

youngersh:
Support user inquire about uadk libraries version with static compilation
like this:
	strings zip_sva_perf | grep version
and can get the static libraries version.

### DIFF
--- a/wd.c
+++ b/wd.c
@@ -12,6 +12,8 @@
 #include "wd.h"
 #include "wd_alg_common.h"
 
+const char *WD_VERSION = "libwd version: 2.3.4";
+
 #define SYS_CLASS_DIR			"/sys/class/uacce"
 
 //wd_log log_out = NULL;


wd.c
```diff
@@ -12,6 +12,8 @@
 #include "wd.h"
 #include "wd_alg_common.h"
 
+const char *WD_VERSION = "libwd version: 2.3.4";
```
hzhuang1, line 15: Why hack the version number at here? The version number is always changing. It's strange to update library version in application.
If user wants to know the library version, he just need to follow the instructions in below.
$ldd zip_sva_perf
This above command will dump the path of libwd.so. Then user could use "ls -l" to display the version number of the library.

hzhuang1: Although the latest UADK library is v2.3.4, user could even run the application with previous UADK library. Since the version number is only for dynamic library. If we hack the version number at here, it'll be totally mess.

youngersh: I know the way to get the version of dynamic library. However, not all users use dynamic libraries to compile applications. In fact, we have no way to get the version of static libraries used by applications unless users  provide it.

wangzhou: The original requirement is we need a version flag or API to get version when compiled as static library. I am not sure this way is a common way, but this is a way we got now. Seems this string maybe not same as dynamic one if they are not updated in once patch.

hzhuang1: Excuse me to response late. Could we do a test with this patch on dynamic library build?
If we run the same command on dynamic library, what's the result? Thanks

youngersh: "string libwd.so.2.3.4 | grep version" will get "libwd version: 2.3.4" too.
In addition, I'm looking for the way to get the version from Makefile.am.

hzhuang1: Do we need to disable it while it's a dynamic library? Leave it only for a static library. Since we don't know which UADK dynamic library is really used by user.

hzhuang1: I do another implementation. Then we could avoid hacking. Please help to review.
https://github.com/Linaro/uadk/pull/263

By the way, it could only dump version from static library. It doesn't match your words that same version number could be dumped from dynamic library.

PR #263, opened 2021-05-01
Version

hzhuang1:


youngersh:
Thanks! I had test the patch and got the expected result. This plan is better than mine.
By the way, 'strings' the dynamic will get the version too.
strings libwd.so | grep version
libwd version: 2.3.5

PR #265, opened 2021-05-08
wd: add wd_get_accel_dev

gaozhangfei:
Get device supporting the algorithm with smallest numa distance
to current numa node

Example:
dev = wd_get_accel_dev("digest");

### DIFF
--- a/wd.c
+++ b/wd.c
@@ -1,4 +1,5 @@
 /* SPDX-License-Identifier: Apache-2.0 */
+#define _GNU_SOURCE
 #include <ctype.h>
 #include <dirent.h>
 #include <errno.h>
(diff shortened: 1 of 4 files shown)


wd.c
```diff
@@ -589,6 +592,34 @@ void wd_free_list_accels(struct uacce_dev_list *list)
...
+	while (list) {
+		tmp = numa_distance(node, list->dev->numa_id);
+		if (dis > tmp && tmp > 0) {
+			dev = list->dev;
+			dis = tmp;
+		}
+		list = list->next;
+	}
+
+	wd_free_list_accels(list);
```
wangzhou, line 618: I think this is wrong as you free the list, so below return dev will be NULL

gaozhangfei: wd_free_list_accels does not free dev

PR #271, opened 2021-05-18
v1: wd_rsa.h remove local header path

gaozhangfei:
LIBRARY_PATH=/usr/local/lib C_INCLUDE_PATH=~/work/uadk/v1 make
uadk/v1/wd_rsa.h:23:19: fatal error: v1/wd.h: No such file or directory

### DIFF
--- a/v1/wd_cipher.c
+++ b/v1/wd_cipher.c
@@ -467,10 +467,17 @@ fail_with_cookies:
 int wcrypto_do_cipher(void *ctx, struct wcrypto_cipher_op_data *opdata,
 		void *tag)
 {
+	int ret;
+
 	if (!tag)
-		return wcrypto_burst_cipher(ctx, &opdata, NULL, 1);
+		ret = wcrypto_burst_cipher(ctx, &opdata, NULL, 1);
 	else
-		return wcrypto_burst_cipher(ctx, &opdata, &tag, 1);
+		ret = wcrypto_burst_cipher(ctx, &opdata, &tag, 1);
+
+	if (likely(ret == 1))
+		return WD_SUCCESS;
+
+	return ret;
 }
 
 int wcrypto_cipher_poll(struct wd_queue *q, unsigned int num)
--- a/v1/wd_rsa.c
+++ b/v1/wd_rsa.c
@@ -923,11 +923,6 @@ static int rsa_request_init(struct wcrypto_rsa_msg *req, struct wcrypto_rsa_op_d
 			WD_ERR("sign or verf in_bytes != key_size!\n");
 			return -WD_EINVAL;
 		}
-
-		if (unlikely(req->out_bytes < c->key_size)) {
-			WD_ERR("out bytes %u error!\n", req->out_bytes);
-			return -WD_EINVAL;
-		}
 	}
 
 	req->key = key;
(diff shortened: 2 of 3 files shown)


hzhuang1:
If user just build UADK v1, and install it into system directory. What's the instructions?

Liulongfang (commented):
Please do not modify the behavior of our interface, if you want to adapt to openssl, please modify the middle layer of the kae engine

gaozhangfei:
> Please do not modify the behavior of our interface, if you want to adapt to openssl, please modify the middle layer of the kae engine

Suppose this is port code error, could you help check with internal code.
Kae is already mature, but different behavior between uacce and warpdrive.
If change kae engine, internal warpdrive can not work

v1/wd_cipher.c
```diff
@@ -385,7 +385,7 @@ static int cipher_recv_sync(struct wcrypto_cipher_ctx *ctx,
 		opdata[i]->status = resp[i]->result;
 	}
 
-	return recv_count;
+	return WD_SUCCESS;
```
Liulongfang, line 388: 这里不改，在下面这里改成这样：
int wcrypto_do_cipher(void *ctx, struct wcrypto_cipher_op_data *opdata,		void *tag)
{
        int ret;
        if (!tag)
                   ret = wcrypto_burst_cipher(ctx, &amp;opdata, NULL, 1);
        else
                   ret = wcrypto_burst_cipher(ctx, &amp;opdata, &amp;tag, 1);
        if (likely(ret == 1))
                   return WD_SUCCESS;
        return ret;
}

v1/wd_rsa.h
```diff
@@ -20,7 +20,7 @@
 #include <stdlib.h>
 #include <errno.h>
 #include <linux/types.h>
-#include "v1/wd.h"
+#include "wd.h"
```
Liulongfang, line 24: You should modify the way of external dependencies, directly changing the path inside will cause compilation errors

v1/wd_rsa.c
```diff
@@ -923,11 +923,6 @@ static int rsa_request_init(struct wcrypto_rsa_msg *req, struct wcrypto_rsa_op_d
 			WD_ERR("sign or verf in_bytes != key_size!\n");
 			return -WD_EINVAL;
 		}
-
-		if (unlikely(req->out_bytes < c->key_size)) {
-			WD_ERR("out bytes %u error!\n", req->out_bytes);
-			return -WD_EINVAL;
-		}
 	}
```
Liulongfang, line 926: It may not work if the output length parameter is not set here。

v1/wd_cipher.c
```diff
@@ -467,10 +467,17 @@ int wcrypto_burst_cipher(void *ctx, struct wcrypto_cipher_op_data **opdata,
...
+		ret = wcrypto_burst_cipher(ctx, &opdata, NULL, 1);
 	else
-		return wcrypto_burst_cipher(ctx, &opdata, &tag, 1);
+		ret = wcrypto_burst_cipher(ctx, &opdata, &tag, 1);
+
+	if (likely(ret == 1))
+		return WD_SUCCESS;
+
+	return ret;
 }
```
Liulongfang, line 481: Here is a complete modification method. I will send a patch tomorrow to fix this problem.

PR #288, opened 2021-06-02
wd: comp: fix logic in wd_do_comp_sync2()

hzhuang1:
When a large buffer is used to compress by wd_do_comp_sync2(), the
condition logic doesn't work. Fix it in wd_do_comp_sync2().

### DIFF
--- a/wd_comp.c
+++ b/wd_comp.c
@@ -438,10 +438,9 @@ int wd_do_comp_sync2(handle_t h_sess, struct wd_comp_req *req)
 {
 	struct wd_comp_sess *sess = (struct wd_comp_sess *)h_sess;
 	struct wd_comp_req strm_req;
-	__u32 total_avail_out = req->dst_len;
 	__u32 chunk = STREAM_CHUNK;
-	__u32 avail_in = 0;
-	__u32 avail_out;
+	__u32 total_avail_in;
+	__u32 total_avail_out;
 	int ret;
 
 	ret = wd_comp_check_params(h_sess, req, CTX_MODE_SYNC);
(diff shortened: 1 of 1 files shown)


wd_comp.c
```diff
@@ -471,20 +470,11 @@ int wd_do_comp_sync2(handle_t h_sess, struct wd_comp_req *req)
...
-		if (req->op_type == WD_DIR_COMPRESS)
-			strm_req.last = (strm_req.src_len == chunk) ? 0 : 1;
-
+		strm_req.src_len = avail_in > chunk ? chunk : avail_in;
 		do {
+			if (req->op_type == WD_DIR_COMPRESS)
+				strm_req.last = (strm_req.src_len == chunk) ? 0 : 1;
 			if (req->op_type == WD_DIR_COMPRESS &&
-			    strm_req.src_len == 0 &&
 			    strm_req.last == 1) {
```
youngersh, line 478: This condition is necessary for the function 'append_store_block'. It only works for the empty block and deals with the 'checksum' or 'isize'.

wd_comp.c
```diff
@@ -471,20 +470,11 @@ int wd_do_comp_sync2(handle_t h_sess, struct wd_comp_req *req)
...
-		avail_in = strm_req.src_len;
-		if (req->op_type == WD_DIR_COMPRESS)
-			strm_req.last = (strm_req.src_len == chunk) ? 0 : 1;
-
+		strm_req.src_len = avail_in > chunk ? chunk : avail_in;
 		do {
+			if (req->op_type == WD_DIR_COMPRESS)
+				strm_req.last = (strm_req.src_len == chunk) ? 0 : 1;
 			if (req->op_type == WD_DIR_COMPRESS &&
-			    strm_req.src_len == 0 &&
```
youngersh, line 487: This condition is necessary for the function 'append_store_block'. It only works for the empty block and deals with the 'checksum' or 'isize'.

wd_comp.c
```diff
@@ -495,8 +485,7 @@ int wd_do_comp_sync2(handle_t h_sess, struct wd_comp_req *req)
 			}
 			dbg("do, strm start, in =%u, out_len =%u\n",
 			    strm_req.src_len, strm_req.dst_len);
-			if (req->dst_len + strm_req.src_len > total_avail_out)
-				return -WD_ENOMEM;
```
haofang111, line 499: test

haofang111: This is just for check if dst space is enough for  hardware to write back. But this judgement  is not correctly maybe occur a false positive. Remove it just as this false positive?

wd_comp.c
```diff
@@ -513,8 +502,8 @@ int wd_do_comp_sync2(handle_t h_sess, struct wd_comp_req *req)
 
 			strm_req.src += strm_req.src_len;
 			avail_in -= strm_req.src_len;
-			strm_req.src_len = avail_in;
-		} while (strm_req.src_len > 0);
+			strm_req.src_len = chunk;
+		} while (avail_in > 0);
```
haofang111, line 507: avail_in meaned single spilt packet avail in size before, but now you change it means stream total avail in.  So there is no need the outer while loop ?

youngersh:
I'm sorry from the comments I can't know which bug you are fixing and how the bug work. If I read this patch right, you want to fix the bug the sync2 return mistakenly. In fact, I think remove line 498 and 499 is enough.

wd_comp.c
haofang111: We just talked about this, the hardware can guarante that not write though memory., remove it is OK!

wd_comp.c
hzhuang1: This sentence doesn't match with the loop condition. The original loop condition is "strm_req.src_len > 0". When "strm_req.src_len == 0", it's already out of the loop.

youngersh: Of course I know this. So the question is what you want to fix. In my opinion, remove line 498/499 is enough if I don't misunderstand your idea.

wd_comp.c
```diff
@@ -464,65 +463,58 @@ int wd_do_comp_sync2(handle_t h_sess, struct wd_comp_req *req)
...
-			req->src_len = 0;
+	do {
+		strm_req.src_len = total_avail_in > chunk ? chunk :
+				   total_avail_in;
+		strm_req.dst_len = total_avail_out > chunk ? chunk :
+				   total_avail_out;
+		if (req->op_type == WD_DIR_COMPRESS) {
+			/* find the last chunk to compress */
+			if (strm_req.src_len < chunk)
+				strm_req.last = 1;
```
haofang111, line 480: It seems can remove this.

wd_comp.c
```diff
@@ -464,65 +463,58 @@ int wd_do_comp_sync2(handle_t h_sess, struct wd_comp_req *req)
...
+	if (req->op_type == WD_DIR_COMPRESS &&
+	    strm_req.src_len == 0 &&
+	    strm_req.last == 1) {
+		dbg("append_store, src_len=%u, dst_len=%u\n",
+		    req->src_len, req->dst_len);
+		ret = append_store_block(h_sess, &strm_req);
+		req->dst_len += strm_req.dst_len;
+		req->status = 0;
+		return 0;
 	}
```
haofang111, line 518: sync2 logic can assurance no empty packet as last，can remove this here.  ShenYang will add this into wd_do_comp_strm becase user can give a empty one.

wd_comp.c
```diff
@@ -464,65 +463,58 @@ int wd_do_comp_sync2(handle_t h_sess, struct wd_comp_req *req)
...
+		    strm_req.src_len, strm_req.dst_len);
+		total_avail_out -= strm_req.dst_len;
 
-			strm_req.src += strm_req.src_len;
-			avail_in -= strm_req.src_len;
-			strm_req.src_len = avail_in;
-		} while (strm_req.src_len > 0);
+		strm_req.src += strm_req.src_len;
+		total_avail_in -= strm_req.src_len;
+		strm_req.src_len = chunk;
```
haofang111, line 501: can remove this line.

hzhuang1:
> Maybe 'total_avail_in <= chunk' is the same?

Yes, it could be same.

wd_comp.c
hzhuang1: > I'm sorry for my previous comments. The append_store_block() is really unreachable. The only question left for me is how to trigger to a bad status 0x41. I agree the opinion you list about the old logic, but I don't understand how it trigger the error. Can you show me the test case? Maybe I need to add it for CI.

It could be reproduced on my branch (hzhuang1/zip_v0.5.7). I'll try to reproduce it on master branch.

wd_comp.c
hzhuang1: Sure.

PR #295, opened 2021-06-11
upstream some bugfix and cleanup

youngersh:


hzhuang1:
> For deflate/zlib/gzip algorithms, it's unnecessary to send a empty last block
> to hardware for stream mode. The deflate algorithm protocol has specified
> the content of an empty tail packet. For zlib and gzip, algorithm protocols
> require adding extra packet tail check data. All these things can be finished
> by software.
>

I'm confused on the comments. It mentioned empty last block is unnecessary. 
But the empty last block is still used in stream mode. They're incoherent.

youngersh:
What I said is sending empty last block to hardware is unnecessary rather than empty last block is unnecessary. The empty can be deal with software.

youngersh:
> > For deflate/zlib/gzip algorithms, it's unnecessary to send a empty last block
> > to hardware for stream mode. The deflate algorithm protocol has specified
> > the content of an empty tail packet. For zlib and gzip, algorithm protocols
> > require adding extra packet tail check data. All these things can be finished
> > by software.
> 
> I'm confused on the comments. It mentioned empty last block is unnecessary.
> But the empty last block is still used in stream mode. They're incoherent.

What I said is sending empty last block to hardware is unnecessary rather than empty last block is unnecessary. The empty can be deal with software.

PR #299, opened 2021-06-16
wd: add environment variable

hzhuang1:
Make UADK to parse environment variable to get information. The basic
goal is to make UADK easier to use.

"--env" is the parameter to enable environment variable in zip_sva_perf
test case. Up to now, the feature of environment variable is enabled in
synchronous mode.

Signed-off-by: Zhou Wang <wangzhou1@hisilicon.com>


Liulongfang (commented):
looks good!

PR #309, opened 2021-06-23
uadk: schedule - modify schedule parameter assignments

hzhuang1:
Now the schedule design three patameters. 'type' is not a common setting.
'numa_id' is a patameter decided by users. So assign it in create 'sess'.

Signed-off-by: Yang Shen <youngershen@foxmail.com>

### DIFF
--- a/include/wd_aead.h
+++ b/include/wd_aead.h
@@ -32,6 +32,7 @@ struct wd_aead_sess_setup {
 	enum wd_cipher_mode cmode;
 	enum wd_digest_type dalg;
 	enum wd_digest_mode dmode;
+	int numa;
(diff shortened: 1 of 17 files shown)


youngersh:
Thx. When I added this patch, I found the 'sched_key' is unused in aead/cipher/digest. So I only fixed the problem in 'hisi_zip_test'.

hzhuang1:
> Thx. When I added this patch, I found the 'sched_key' is unused in aead/cipher/digest. So I only fixed the problem in 'hisi_zip_test'.

Please help to submit your PR with your new additional patch. Let's check the status now.

PR #310, opened 2021-06-23
wd: add cipher environment variable

lin755:
1.Adapt cipher to init ctx with environment variable
2."--use_env" for test_hisi_sec to enable environment variable.
3.Fix some problem under default scenario.

More algorithm modules will be adapted to use environment variable  later.

### DIFF
--- a/wd_cipher.c
+++ b/wd_cipher.c
@@ -18,6 +18,7 @@
 #define WD_POOL_MAX_ENTRIES	1024
 #define DES_WEAK_KEY_NUM	4
 #define MAX_RETRY_COUNTS	200000000
+#define WD_CIPHER_ENV_NUM	4
 
 #define POLL_SIZE		1000000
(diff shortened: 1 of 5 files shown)


wd_cipher.c
```diff
@@ -415,6 +418,9 @@ int wd_do_cipher_sync(handle_t h_sess, struct wd_cipher_req *req)
 	} while (ret < 0);
 	pthread_spin_unlock(&ctx->lock);
 
+	if (wd_cipher_env_config.enable_internal_poll)
+		wd_add_task_to_async_queue(&wd_cipher_env_config, idx);
+
```
hzhuang1, line 423: I don't understand why add it into async queue while you're running in sync mode. Do you write it in the wrong place?

lin755: It is a mistake when sync code,I have changed it

PR #338, opened 2021-07-12
uadk/digest - add stream mode for digest

yekai123123:
1. Add stream mode for digest
2. modify the stream mode test


gaozhangfei (commented):
I prefer directly using wd_get_accel_dev in uadk_engine_digests, it's simple enough, no need add new api, as well as new macro.
Only my view.

hzhuang1:
Need to rebase and replay zhangfei's comment first. So close it.

PR #339, opened 2021-07-12
wd env: add some api

lin755:
1.wd_alg_ctx_num is used if system not support env set
2.wd_get_env_param is used to get specificed ctx num


wd_comp.c
```diff
@@ -778,6 +778,7 @@ static const struct wd_alg_ops wd_comp_ops = {
 
 int wd_comp_env_init(void)
 {
+	wd_comp_env_config.disable_env = 0;
```
hzhuang1, line 781: What's the meaning of "disable_env" at here? Why do you need to clear it and set it again?

lin755: I use it  to split 2 scenarios here, 
if (config->disable_env)
    var_s = var->def_val;//If we use wd_alg_ctx_num ,we do not want system env influence
else
    var_s = getenv(var->name);//get value from system env

wd_comp.c
```diff
@@ -786,3 +787,61 @@ void wd_comp_env_uninit(void)
 {
 	return wd_alg_env_uninit(&wd_comp_env_config);
 }
+
+int wd_comp_ctx_num(__u32 node, __u32 type, __u32 num, __u8 mode)
```
hzhuang1, line 791: Which function is invoking this function? I can't any function reference to it.

PR #340, opened 2021-07-14
Env v0.3

hzhuang1:


wd_util.c
```diff
@@ -351,34 +352,72 @@ int wd_parse_async_poll_en(struct wd_env_config *config, const char *s)
 	return 0;
 }
 
-/* fix me */
-static int parse_ctx_num_on_numa(const char *s, int *ctx_num, int *node)
+static int parse_num_on_numa(const char *s, int *num, int *node)
```
lin755, line 355: I have add a number check in this function

hzhuang1: The number check should be common. Is it right?

lin755 (approved):
LGTM.

PR #344, opened 2021-08-02
Optimize for wd env and code clean

lin755:
use WD_<alg>_CTX_NUM to define alg ctx number.
for zip,use by WD_COMP_CTX_NUM=""sync-comp:1@0,async-decomp:1@0"
for sec and hpre,use by WD_COMP_CTX_NUM=""sync:1@0,async:1@0"

### DIFF
--- a/drv/hisi_comp.c
+++ b/drv/hisi_comp.c
@@ -855,7 +855,7 @@ static int hisi_zip_init(struct wd_ctx_config_internal *config, void *priv)
 	handle_t h_ctx;
 	int i;
 
-	memcpy(&zip_ctx->config, config, sizeof(struct wd_ctx_config));
+	memcpy(&zip_ctx->config, config, sizeof(struct wd_ctx_config_internal));
 	/* allocate qp for each context */
 	for (i = 0; i < config->ctx_num; i++) {
 		h_ctx = config->ctxs[i].ctx;
--- a/test/hisi_zip_test/testsuit.c
+++ b/test/hisi_zip_test/testsuit.c
@@ -805,6 +805,8 @@ int test_hw(struct test_options *opts, char *model)
 	struct stat statbuf;
 	chunk_list_t *tlist = NULL;
 	int div;
+	__u32 num;
+	__u8 enable;
 
 	if (!opts || !model) {
 		ret = -EINVAL;
@@ -898,6 +900,14 @@ int test_hw(struct test_options *opts, char *model)
 		ret = nonenv_resource_init(opts, &info, &sched);
 	if (ret < 0)
 		goto out;
+
+	if (opts->use_env) {
+		ret = wd_comp_get_env_param(0, opts->op_type, opts->sync_mode, &num, &enable);
+		printf("get_env_param ret:%d, ctx num:%u, enable poll:%u",
+		       ret, num, enable);
+		if (ret < 0)
+			goto out;
+	}
 
 	if (opts->is_file) {
 		ret = fstat(opts->fd_in, &statbuf);
--- a/wd_comp.c
+++ b/wd_comp.c
@@ -20,8 +20,6 @@
 #define POLL_SIZE			250000
 #define POLL_TIME			1000
 
-#define WD_ARRAY_SIZE(array)           (sizeof(array) / sizeof(array[0]))
-
 #define swap_byte(x) \
 	((((x) & 0x000000ff) << 24) | \
 	(((x) & 0x0000ff00) <<  8) | \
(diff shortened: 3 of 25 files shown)


wd_util.c
```diff
@@ -1092,44 +1109,183 @@ static void wd_uninit_resource(struct wd_env_config *config)
...
+	if (mode >= CTX_MODE_MAX) {
+		WD_ERR("wrong ctx mode(%u))!\n", mode);
+		return -WD_EINVAL;
+	}
+
+	ctx_attr->node = node;
+	ctx_attr->mode = mode;
+	ctx_attr->num = num;
+	/* If type is CTX_TYPE_INVALID, we need update it to 0. */
+	ctx_attr->type = (type == CTX_TYPE_INVALID) ? 0 : type;
```
hzhuang1, line 1286: If mode parameter is invalid, assignment isn't executed. Now type parameter is invalid too, it seems that we should skip the assignment.

wd_comp.c
```diff
@@ -768,10 +759,52 @@ static const struct wd_alg_ops wd_comp_ops = {
...
+			       &wd_comp_ops, ARRAY_SIZE(table), &ctx_attr);
+}
+
+void wd_comp_ctx_num_uninit(void)
+{
+	return wd_alg_env_uninit(&wd_comp_env_config);
+}
+
+int wd_comp_get_env_param(__u32 node, __u32 type, __u32 mode,
+			  __u32 *num, __u8 *is_enable)
```
hzhuang1, line 794: How to use this function? I can't find any code reference to this function.

lin755: We could use like this, to get decomp async ctx num on node 0.
__u32 num;
__u8 enable;
ret = wd_comp_get_env_param(0, 1, 1, &num, &enable)

wd_util.c
lin755: type is set to 0, so we can use ctx_table[mode][type] also for cipher，rsa etc

wd_comp.c
hzhuang1: So it should be invoked by wd_comp driver. Don't add any code without reference. You could append a new patch to use this function.

drv/hisi_comp.c
```diff
@@ -855,7 +855,7 @@ static int hisi_zip_init(struct wd_ctx_config_internal *config, void *priv)
 	handle_t h_ctx;
 	int i;
 
-	memcpy(&zip_ctx->config, config, sizeof(struct wd_ctx_config));
+	memcpy(&zip_ctx->config, config, sizeof(struct wd_ctx_config_internal));
```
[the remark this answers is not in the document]
lin755: If wd_ctx_config_internal is changed , like add some struct member, it will casuse  segment fault when hisi_zip_exit

wd_comp.c
lin755: ok,I will add it in test file

drv/hisi_comp.c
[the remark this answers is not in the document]
hzhuang1: If more struct members are appended in internal config, your instruction will only copy chunk data into internal config. If some struct members are removed from internal config (the size of internal config is less than the size of config), your instruction will trigger segment fault since some extra data is copied.

If more struct members are appended in internal config, original instruction will only copy limited data. If some struct members are removed from internal config (the size of internal config is less than the size of config), original instructionn will trigger segment fault since some extra data is copied.

So I suggest to add an extra check whether the size is OK.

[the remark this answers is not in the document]
lin755: struct hisi_zip_ctx {
        struct wd_ctx_config_internal   config;
};

I found that zip_ctx->config is also type of wd_ctx_internal_config，so  there is a special reason to copy that sizeof wd_ctx_config? This memcpy is also used in sec and hpre init, I check their size is also sizeof(struct wd_ctx_config_internal)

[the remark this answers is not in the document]
hzhuang1: Could we change it to "struct wd_ctx_config"? If it's an internal structure, it shouldn't be accessed by test cases. It should only be accessed by UADK internally.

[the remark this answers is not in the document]
lin755: wd_ctx_config is set by user, and wd_ctx_config_internal is used in side uadk, it is copy  from wd_ctx_config  when wd_comp_init, but two struct may be different, so I think use internal struct in driver level is reasonable.

test/hisi_zip_test/testsuit.c
```diff
@@ -899,6 +901,14 @@ int test_hw(struct test_options *opts, char *model)
 	if (ret < 0)
 		goto out;
 
+	if (opts->use_env) {
+		ret = wd_comp_get_env_param(0, 0, 0, &num, &enable);
```
hzhuang1, line 905: I'm curious on this API. It seems that ctx & enable are not used at all. If so, why do we need to add these two parameters in this API?

Do you have any plan to use these two parameters?

By the way, you force the numa, mode and type parameter in test case. At least, mode and type are depend on cases.

lin755: They are designed to save ctx num and flag,it is a dfx function to user.

I have changed  mode and type parameter to opts op_type and sync_mode now.

lin755: I discuss with tanghui, it seems ctx num and enable flag is needed by openssl engine async mode.

PR #352, opened 2021-08-27
wd: sec: remove ctx_num limitation

gaozhangfei:
Nginx doest fork worker processes after master process
So remove the limitaiton here to make fork possible

### DIFF
--- a/wd_cipher.c
+++ b/wd_cipher.c
@@ -206,11 +206,6 @@ int wd_cipher_init(struct wd_ctx_config *config, struct wd_sched *sched)
 	void *priv;
 	int ret;
 
-	if (wd_cipher_setting.config.ctx_num) {
-		WD_ERR("cipher have initialized.\n");
-		return -WD_EEXIST;
-	}
-
 	if (!config || !sched) {
 		WD_ERR("wd cipher config or sched is NULL!\n");
 		return -WD_EINVAL;
--- a/wd_digest.c
+++ b/wd_digest.c
@@ -133,11 +133,6 @@ int wd_digest_init(struct wd_ctx_config *config, struct wd_sched *sched)
 	void *priv;
 	int ret;
 
-	if (wd_digest_setting.config.ctx_num) {
-		WD_ERR("digest have initialized.\n");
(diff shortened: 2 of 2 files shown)


hzhuang1:
@Liulongfang Please help to review this pull request.

Liulongfang (commented):
After removing it, how to prevent repeated initialization? Repeated initialization will lead to resource leakage

gaozhangfei:
1. uadk is a library, which should not prevent an application double initializes.
 Instead it is applications' responsibility.
for example, glibc does not prevent application double malloc without free

2. It's incorrect to use a static variable prevent double initialization.
Considering multi-process.
Parent init, static a = true.
fork
child can not init again since static a == true.
Is this expected?

This is why nginx fail to boot, since digest and cipher are used in master process to init.

gaozhangfei:
1. uadk is a library, which should not prevent an application double initializes.
 Instead it is applications' responsibility.
for example, glibc does not prevent application double malloc without free

2. It's incorrect to use a static variable prevent double initialization.
Considering multi-process.
Parent init, static a = true.
fork
child can not init again since static a == true.
Is this expected?

This is why nginx fail to boot, since digest and cipher are used in master process to init.

Liulongfang:
What you said is reasonable, but your modification method did not solve the problem you mentioned, and this modification will lead to resource leakage.

gaozhangfei:
how cause resource leakage?

Liulongfang:
You can look at the wd_cipher_uninit(), all the free resources inside have been leaked.

gaozhangfei:
void wd_cipher_uninit(void)
{
        void *priv = wd_cipher_setting.priv;

        if (!priv)
                return;

        wd_cipher_setting.driver->exit(priv);
        wd_cipher_setting.priv = NULL;
        free(priv);

        wd_uninit_async_request_pool(&wd_cipher_setting.pool);
        wd_clear_sched(&wd_cipher_setting.sched);
        wd_clear_ctx_config(&wd_cipher_setting.config);
}

any relation with wd_cipher_setting.config.ctx_num?

PR #358, opened 2021-08-31
wd: avoid to install la files

hzhuang1:
la files are generated in both static library or dynamic library build.
In different build, la files are different. But la files are always
installed into library directory. It will confuse test cases.

Now delete la files before installing libraries.


Liulongfang (commented):
OK

PR #365, opened 2021-09-08
uadk/digest - add stream mode for digest sync mode

yekai123123:
Add the sec digest state for steam mode. User can use it to
notify the digest state. Using the digest out memory to store the
total of packet, So user should allocated the memory with at
least twice the digest length.


yekai123123:
need to update a new code.

PR #366, opened 2021-09-09
uadk/benchmark: modify test tools data cal

Liulongfang:
Remove the alignment operation of the number of received and
send packets in the uadk benchmark, and calculate the performance
according to the time stop method.

Reduce the impact of test tool code on performance by optimizing
the statistical method of test tools.

### DIFF
--- a/uadk_benchmark/sec_uadk_benchmark.c
+++ b/uadk_benchmark/sec_uadk_benchmark.c
@@ -546,8 +546,8 @@ void *sec_uadk_poll(void *data)
 	typedef int (*poll_ctx)(__u32 idx, __u32 expt, __u32 *count);
 	poll_ctx uadk_poll_ctx = NULL;
 	thread_data *pdata = (thread_data *)data;
-	u32 send_time, send_num;
-	u32 expt = 0;
+	u32 expt = ACC_QUEUE_SIZE * g_thread_num;
+	u32 last_time = 2; /* poll need one more recv time */
 	u32 count = 0;
 	u32 recv = 0;
 	u32 i = 0;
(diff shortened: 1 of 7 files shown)


uadk_benchmark/sec_uadk_benchmark.c
```diff
@@ -546,8 +546,8 @@ void *sec_uadk_poll(void *data)
 	typedef int (*poll_ctx)(__u32 idx, __u32 expt, __u32 *count);
 	poll_ctx uadk_poll_ctx = NULL;
 	thread_data *pdata = (thread_data *)data;
-	u32 send_time, send_num;
-	u32 expt = 0;
+	u32 expt = ACC_QUEUE_SIZE * g_thread_num;
+	u32 last_time = 2; /* poll need one more recv time */
```
hzhuang1, line 550: You mentioned that poll need one more recv time. But I don't understand it. Could you help to share more comments?

Liulongfang: The message is sent during the last round of the timer check. At this time, if the poll thread also happens to check the timer status and the time just expires, the message sent by the sending thread has not been executed by the hardware yet, so an extra one is needed. Polls are used to recover the last sent message to obtain more accurate performance data.
If the message sent is very large, which leads to a long hardware execution time, then the last message sent will not be received by poll。
 if the time exceeds too much at that time, and the performance within the specified time should not be calculated.

hzhuang1: Why not just check whether all sending packets are all received? If all sending packets are received, polling thread could exist. Otherwise, polling thread should keep polling.

Liulongfang: Because of the comparison of the received quantity, it will cause the asynchronous transmission and reception to be submitted every time it is sent, and the receiving end also needs to receive the quantity to be submitted. If it is sent in multiple threads or received in multiple threads, this comparison will be caused. And submission becomes a hot function, which will affect the efficiency of its own business receiving and sending

PR #370, opened 2021-09-13
wd env bugfix and some clean code

lin755:


wd_util.c
```diff
@@ -312,12 +312,76 @@ void dump_env_info(struct wd_env_config *config)
...
+		return;
+
+	FOREACH_NUMA(i, config, config_numa)
+		free(config_numa->dev);
+
+	free(config->config_per_numa);
+}
+
+static void wd_get_dev_numa(struct uacce_dev_list *head,
+			    int *numa_dev_num, int *numa_num)
```
hzhuang1, line 345: Could you help to share more comments on **numa_dev_num** and **numa_num**?

wd_util.c
```diff
@@ -312,12 +312,76 @@ void dump_env_info(struct wd_env_config *config)
...
+{
+	struct uacce_dev_list *list = head;
+
+	while (list) {
+		if (list->dev->numa_id < 0)
+			list->dev->numa_id = 0;
+
+		if (!numa_dev_num[list->dev->numa_id])
+			(*numa_num)++;
```
hzhuang1, line 354: You increased **\*numa_num** directly. But what's the initial/expected value of **\*numa_num**?

wd_util.c
lin755: numa_dev_num represents how many device of one type(like sec2 support cipher) on this numa, numa_num represents how many numa has such devices. 
For example, if we are on kunpeng 920, only node 0 and node 2 has sec2 devices,but node 0 may connect with two  sec2 devices,so numa_dev_num is 2, not 1.

hzhuang1: Good. Please comment it in the code.

wd_util.c
lin755: You can see numa_num initial value is 0 in wd_alloc_numa statement. Add it one time only when we find a device on a new numa node.if device numa node is 0, 0, 2，we add it one only in first and third device

wd_util.c
lin755: ok,I will change it later

wd_util.c
hzhuang1: It seems that wd_get_dev_numa() is only called once. And **\*numa_num** is initialized out of the wd_get_dev_numa(). Maybe you can initialize it just in wd_get_dev_numa().

lin755: Resubmit these changes now.

PR #371, opened 2021-09-13
wd: bugfix QM's barriers and spin lock

Liulongfang:
the original __sync_synchronize() is a general memory synchronization
operation, and rmb()/wmb() will be distinguished according to the type
of operation, thereby reducing the blocking time and improving
performance

the range of the spin lock should be as small as possible to reduce the
performance impact.


hzhuang1:
Could you help to share data how much is improved with this patch?

Liulongfang:
Improved performance by about 5%

PR #374, opened 2021-09-17
wd: alg: fix poll ctx

lin755:
Count may be a dirty data, so init it first.

Signed-off-by: Wenkai Lin <linwenkai6@hisilicon.com>


gaozhangfei:
Test OK

PR #375, opened 2021-09-22
cipher: fix wrong iv length check

youngersh:
There is no 'iv' in XTS mode, so there is no need to verify
the 'iv' length for XTS mode.

Signed-off-by: Yang Shen <shenyang39@huawei.com>

### DIFF
--- a/wd_cipher.c
+++ b/wd_cipher.c
@@ -309,7 +309,7 @@ static int cipher_iv_len_check(struct wd_cipher_req *req,
 {
 	int ret = 0;
 
-	if (sess->mode == WD_CIPHER_ECB)
+	if (sess->mode == WD_CIPHER_ECB || sess->mode == WD_CIPHER_XTS)
 		return 0;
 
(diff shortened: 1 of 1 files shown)


gaozhangfei:
do we need update src/uadk_cipher.c as well

youngersh:
> do we need update src/uadk_cipher.c as well

Update what?

gaozhangfei:
https://github.com/Linaro/openssl-uadk/pull/65/commits/82769704f53188d492e830e5d45d072f99a7b71c
no xts para change in this commit

youngersh:
> [Linaro/openssl-uadk@8276970](https://github.com/Linaro/openssl-uadk/commit/82769704f53188d492e830e5d45d072f99a7b71c)
> no xts para change in this commit

I have add this patch to pr #65 and fix the xts iv params.

gaozhangfei:
ok, I see

Tested OK

PR #378, opened 2021-09-28
wd: bugfix msg pool idx

Liulongfang:
1.reduce the number of queries to find effective resources from the
resource pool by retaining the index of the effective resource
obtained last time.
2.bugfix some coding style

### DIFF
--- a/wd_util.c
+++ b/wd_util.c
@@ -214,8 +214,8 @@ void wd_uninit_async_request_pool(struct wd_async_msg_pool *pool)
 /* fix me: this is old wd_get_req_from_pool */
 void *wd_find_msg_in_pool(struct wd_async_msg_pool *pool, int index, __u32 tag)
 {
-	struct msg_pool *p;
-	__u32 msg_num = pool->pools[index].msg_num;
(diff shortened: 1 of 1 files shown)


wd_util.c
```diff
@@ -246,24 +241,23 @@ int wd_get_msg_from_pool(struct wd_async_msg_pool *pool, int index, void **msg)
 			return -WD_EBUSY;
 	}
 
+	p->tail = (idx + 1) % msg_num;
```
hzhuang1, line 244: 1. **head** & **tail** were both defined without usage. Now only **tail** is enabled in this patch. But **head** is still no used. **head** and **tail** are always in pair. I suggest to rename it to **last**. If **tail** is used, user always think where is **head** and how to use it.
2. It's not good to update the **tail** directly. The assignment is composed by a few instructions. It could be switched to other threads at any time. I think that we could use atomic store on updating **tail**. Same on loading **tail** above.

hzhuang1: **tail** is also fine. Please help to remove **head** since it's not used at all.

PR #382, opened 2021-10-20
wd: sec: change init entry limitation per pid

gaozhangfei:
Nginx master process do the init, then fork worker
processes do the real transaction. So change the
entry limitaiton according to pid.


Liulongfang:
I put this part of the modification to wd_util.c, using a public modification method

PR #383, opened 2021-10-21
wd: bugfix multi-process initialization function

Liulongfang:
UADK device initialization is based on the process. if multiple
initialization occur in a process, it will cause resource leakage.
when Nginx master creates a new process through the fork process,
some parameters will be copied, but the real device resource are not
initialized.
Therefore, It need to distinguish based on pid whethre device
initialization is required.

### DIFF
--- a/wd_util.c
+++ b/wd_util.c
@@ -91,6 +91,7 @@ int wd_init_ctx_config(struct wd_ctx_config_internal *in,
 	}
 
 	in->ctxs = ctxs;
+	in->pid = getpid();
 	in->priv = cfg->priv;
 	in->ctx_num = cfg->ctx_num;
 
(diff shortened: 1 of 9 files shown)


wd_util.c
```diff
@@ -91,6 +91,7 @@ int wd_init_ctx_config(struct wd_ctx_config_internal *in,
 	}
```
gaozhangfei, line 91: if wd_init_ctx_config is used by all algorithms, we can simply add pid check inside wd_init_ctx_config instead of every algorithms.c
wd_init_ctx_config
   if (in->pid == getpid)
    return;

Liulongfang: Yes，this way is better.

PR #388, opened 2021-10-30
drv/sec: refactoring the sec user space code

yekai123123:
1. Modify the code to reduced repetition rate
2. remove some redundant code

### DIFF
--- a/drv/hisi_sec.c
+++ b/drv/hisi_sec.c
@@ -531,7 +531,6 @@ static void update_iv(struct wd_cipher_msg *msg)
 				msg->iv_bytes, msg->iv_bytes);
 		break;
 	case WD_CIPHER_CTR:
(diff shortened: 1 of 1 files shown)


drv/hisi_sec.c
```diff
@@ -531,7 +531,6 @@ static void update_iv(struct wd_cipher_msg *msg)
 				msg->iv_bytes, msg->iv_bytes);
 		break;
 	case WD_CIPHER_CTR:
-		if (msg->iv_bytes >= AES_BLOCK_SIZE)
```
hzhuang1, line 534: You didn't explain why the conditional code should be removed.

yekai123123: Due to has added the iv length checking in wd_cipher. so the ctr mode iv length checking is redundant.

PR #405, opened 2021-11-24
wd/v1: fix get_dev_info return value

lin755:
get_dev_info should not return flags, it is meaningless.

Signed-off-by: Wenkai Lin <linwenkai6@hisilicon.com>

### DIFF
--- a/v1/wd.c
+++ b/v1/wd.c
@@ -258,8 +258,11 @@ static int get_int_attr_all(struct dev_info *dinfo)
 	dinfo->node_id = get_int_attr(dinfo, "node_id");
 
 	ret = get_int_attr(dinfo, "flags");
-	if (ret < 0 || (unsigned int)ret & WD_UACCE_DEV_SVA)
+	if (ret < 0)
 		return ret;
+	else if ((unsigned int)ret & WD_UACCE_DEV_SVA)
+		return -ENODEV;
+
 	dinfo->flags = ret;
 
 	return 0;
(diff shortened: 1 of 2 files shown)


v1/wd.c
```diff
@@ -258,8 +258,11 @@ static int get_int_attr_all(struct dev_info *dinfo)
 	dinfo->node_id = get_int_attr(dinfo, "node_id");
 
 	ret = get_int_attr(dinfo, "flags");
-	if (ret < 0 || (unsigned int)ret & WD_UACCE_DEV_SVA)
+	if (ret < 0)
 		return ret;
+	else if ((unsigned int)ret & WD_UACCE_DEV_SVA)
+		return -ENODEV;
+
```
hzhuang1, line 265: According to your comment, checking flag is unnecessary since there's no flag on return value. Then why do you still check it? Why not delete it at all?

lin755: If the device is  SVA，we should skip it by ENODEV.
If we not change it, it will return an number bigger than 0(like 129 read from uacce flags), the last caller wd_get_available_dev_num may take it as a normal device number.

lin755: PS，wd_get_available_dev_num is a no-sva api，so we need skip sva mode device。

hzhuang1:
You still didn’t answer why comments are not in line with the code.

You code just re-arrange the conditional code without changing anything. Maybe you need to update the comments with “Avoid too much conditional code in one line.”.

lin755:
Its logical has changed.
before:
If ret < 0, return ret, 
if ret >=0 and ret & SVA,**return ret.**
if ret >=0 and !(ret & SVA) ,return 0 .
after:
If ret < 0, return ret, 
if ret >=0 and ret & SVA,**return -ENODEV**
if ret >=0 and !(ret & SVA) ,return 0 .
Anyway, I will change commit message later.

PR #406, opened 2021-11-27
wd/sched: update the RR scheduler inside UADK

Liulongfang:
Since the current default scheduling method will have a serious
performance bottleneck under multi-threading, it is necessary
to update the scheduling implementation method.

1.Update the interface of the RR scheduler
2.Update the usage of scheduler in ALG API
3.Update the uadk test tools


Liulongfang:
need more modify

PR #416, opened 2021-12-09
supports digest stream mode

yekai123123:


gaozhangfei:
this caused nginx fail to work
uadk master + openssl-uadk master NOT work on nginx

PR #449, opened 2022-02-23
uadk: update file permission on cleanup

hzhuang1:
Set execution permission on cleanup.sh.


hzhuang1:
The pull request can't be verified by CI. Close and create a new one instead.

PR #467, opened 2022-04-02
wd/sched: modify default sched param

Liulongfang:
In the general scenario, after the scheduler is initialized, the user
wished to use the default scheduling parameters through the default
configuration when initilazing the task session

Signed-off-by: Liulongfang <liulongfang@foxmail.com>

### DIFF
--- a/wd_sched.c
+++ b/wd_sched.c
@@ -4,8 +4,10 @@
  * Copyright 2020-2021 Linaro ltd.
  */
 
+#define _GNU_SOURCE
 #include <stdlib.h>
 #include <stdbool.h>
+#include <sched.h>
 #include <numa.h>
 #include "wd_sched.h"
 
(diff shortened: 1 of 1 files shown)


wd_sched.c
```diff
@@ -117,8 +120,10 @@ static struct sched_ctx_region *sched_get_ctx_range(struct wd_sched_ctx *ctx,
 
 	/* If the key->numa_id is not exist, we should scan for a region */
 	for (numa_id = 0; numa_id < ctx->numa_num; numa_id++) {
-		if (sched_info[numa_id].ctx_region[key->mode][key->type].valid)
-			return &sched_info[numa_id].ctx_region[key->mode][key->type];
+		for (type = 0; type < ctx->type_num; type++) {
+			if (sched_info[numa_id].ctx_region[key->mode][type].valid)
+				return &sched_info[numa_id].ctx_region[key->mode][type];
+		}
```
hzhuang1, line 126: Why do you change from **key->type** to **type**?

Liulongfang: In line 119, the scene where the key has been filled with parameters has been processed. 
In fact, the parameters in the key are all default, and a valid region can only be found by querying

hzhuang1: Your assumption is that there's only one **type** in the ctx_region. If there're multiple **type**s in the ctx_region, what will happen?

Liulongfang: If there are multiple types in ctx_region, when there are input parameters, the region can be obtained directly through the parameters. If there are no input parameters, a region in the nearest numa field is returned for the user to use. If the types do not match, an error will be reported in the uadk driver

PR #486, opened 2022-07-21
fix qm and sgl lock

qwl731:
1. Fix missing destroy pool lock and q_info lock.
2. split send and recv BD q_info lock.


Liulongfang (approved):
OK

Liulongfang (approved):

PR #492, opened 2022-08-15
cipher/digest: supports new AES mode algs

yekai123123:
include cipher and digest algs.


drv/hisi_sec.c
```diff
@@ -1566,6 +1588,13 @@ int hisi_sec_digest_send(handle_t ctx, void *digest_msg)
 	if (ret)
 		goto put_sgl;
 
+	/*
+	 *     [has_next , iv_bytes]
+	 *     [    1    ,     0   ]   =   long hash(frist bd)
+	 *     [    1    ,     1   ]   =   long hash(middle bd)
+	 *     [    0    ,     1   ]   =   long hash(end bd)
+	 *     [    0    ,     0   ]   =   block hash(single bd)
+	 */
```
Liulongfang, line 1597: The third parameter mode cannot distinguish whether it is the last tail bd of the long hash mode or the abnormal BD of the block mode

Liulongfang (approved):
OK

drv/hisi_sec.c
yekai123123: fix it in the next patch, have deleted this patch.

PR #493, opened 2022-08-22
uadk - add wd_comp_init2

youngersh:
The current uadk initialization process is:
1.Call wd_request_ctx() to request ctxs from devices.
2.Call wd_sched_rr_alloc() to create a sched(or some other scheduler alloc
  function if exits).
3.Initialize the sched.
4.Call wd_<alg>_init() with ctx_config and sched.

Logic is reasonable. But in practice, the step of `wd_ request_ Ctx() `
and `wd_ sched_ rr_alloc() ` are very tedious. This makes it difficult
for users to use the interface. One of the main reasons for this is
that uadk has made a lot of configurations in the scheduler in order
to provide users with better performance. Based on this consideration,
the current uadk requires the user to arrange the division of hardware
resources according to the device topology during initialization.
Therefore, as a high-level interface, this scheme can provide customized
scheme configuration for users with deep needs.

All algorithm initialization interfaces have the same input parameters
and behavioral logic. The pre-processing of the wd_<alg>_init is actually
the configuration of `struct wd_ctx_config` and `struct wd_sched`.
Therefore, the next thing to be done is to use limited and easy-to-use
input parameters to describe users' requirements on the two input
parameters, ensuring that the functions of the new interface init2
are the same as those of init. For ease of description, v1 is used
to refer to the existing interface, and v2 is used to refer to the layer
of encapsulation.

At present, at least 4 parameters are required to meet the user
configuration requirements with the V1 interface function remains
unchanged.
@device_list: The available uacce device list. Users can get it by
wd_get_accel_list().
@numa_bitmask: The bitmask provided by libnuma. Users can use this
parameter to control requesting ctxs devices in the bind NUMA scenario.
@ctx_nums: The requested ctx number for each numa node. Due to users
may have different requirements for different types of ctx numbers,
needs a two-dimensional array as input.
@sched_type: Scheduling type the user wants to use.

What's more, some users want uadk to provide the default value about
input parameters for some performance insensitive scenes. C code has
no way to.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -87,7 +87,7 @@ AM_CFLAGS += -DWD_NO_LOG
 
 libwd_la_LIBADD = $(libwd_la_OBJECTS) -lnuma
 
-libwd_comp_la_LIBADD = $(libwd_la_OBJECTS) -ldl
+libwd_comp_la_LIBADD = $(libwd_la_OBJECTS) -ldl -lnuma
 libwd_comp_la_DEPENDENCIES = libwd.la
 
 libhisi_zip_la_LIBADD = -ldl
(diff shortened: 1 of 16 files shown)


hzhuang1:
While your patch is applied, does it impact the performance?

PR #494, opened 2022-08-22
Sched

youngersh:
Now the uadk scheduler will search the nearest node when call
session_sched_init(). But this is a repetitive work because
session_sched_init() is called by wd_xxx_create_sess().

So we can build an cpu-device numa node map. Update the map when
a new numa node has been enabled with a device. And then,
session_sched_init() can quick get the right device id from
a certain cpu.

### DIFF
--- a/wd_sched.c
+++ b/wd_sched.c
@@ -70,13 +70,15 @@ struct wd_sched_info {
  * @numa_num: the max numa numbers of the scheduler.
  * @type_num: the max operation types of the scheduler.
  * @poll_func: the task's poll operation function.
- * @sched_info: the context of the scheduler
+ * @numa_map: a map of cpus to devices.
+ * @sched_info: the context of the scheduler.
  */
 struct wd_sched_ctx {
 	__u32 policy;
 	__u32 type_num;
 	__u16  numa_num;
 	user_poll_func poll_func;
+	int numa_map[NUMA_NUM_NODES];
 	struct wd_sched_info sched_info[0];
 };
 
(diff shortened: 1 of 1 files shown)


wd_sched.c
```diff
@@ -501,8 +496,7 @@ struct wd_sched *wd_sched_rr_alloc(__u8 sched_type, __u8 type_num,
 		return NULL;
 
 	if (sched_type >= SCHED_POLICY_BUTT || !type_num) {
-		WD_ERR("invalid: sched_type is %u or type_num is %u!\n",
-		       sched_type, type_num);
+		WD_ERR("invalid: sched_type is %u or type_num is %u!\n", sched_type, type_num);
 		return NULL;
 	}
```
Liulongfang, line 522: Now that the scheduler has been modified, add this missing assignment as well:
sched->sched_policy = sched_type;
sched->name = sched_table[sched_type].name;

wd_sched.c
```diff
@@ -429,6 +447,8 @@ int wd_sched_rr_instance(const struct wd_sched *sched, struct sched_params *para
 	sched_info[numa_id].ctx_region[mode][type].valid = true;
 	sched_info[numa_id].valid = true;
 
+	wd_sched_map_cpus_to_dev(sched_ctx);
```
Liulongfang, line 450: It is recommended that you test the scenario in which the scheduler is used to execute business on a multi-device platform. Your logic here will cause errors.

wd_sched.c
youngersh: Got it and I'll update this next version.
Thx!

PR #495, opened 2022-08-27
uadk: Update the symbol list for uadk libraries

Liulongfang:
Explicitly declare the function symbol list of its
external interface for the so dynamic library file of uadk,
only make the external interface functions visible to the
outside APP, and all other functions are only used internally.

Signed-off-by: liulongfang <liulongfang@huawei.com>


haofang111:
looks good to me！

PR #498, opened 2022-09-08
wd/sched: fix skey->type when param->numa_id < 0

gaozhangfei:
When param->numa_id < 0, skey->type is always 0.
Fixed by setting the type in both cases.

### DIFF
--- a/wd_sched.c
+++ b/wd_sched.c
@@ -197,9 +197,11 @@ static handle_t session_sched_init(handle_t h_sched_ctx, void *sched_param)
 			return (handle_t)(-WD_ENOMEM);
 		}
 	} else {
-		skey->type = param->type;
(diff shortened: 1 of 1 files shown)


wd_sched.c
```diff
@@ -197,10 +197,12 @@ static handle_t session_sched_init(handle_t h_sched_ctx, void *sched_param)
 			return (handle_t)(-WD_ENOMEM);
 		}
 	} else {
-		skey->type = param->type;
 		skey->numa_id = param->numa_id;
 	}
 
+	if (param)
+		skey->type = param->type;
```
Liulongfang, line 204: Your modification here is not complete. I have a patch that needs to be added, and I will deal with your modification together.

PR #499, opened 2022-09-13
uadk/sched: modify the automatic scheduling function

Liulongfang:
In the scenario of setting the default numa parameters, use the numa affinity method to obtain the numa scheduling domain, so as to achieve automatic balanced use of hardware resources.

Signed-off-by: Longfang Liu <liulongfang@huawei.com>


gaozhangfei:
Test OK, thanks

PR #503, opened 2022-09-23
Optimized device query method

Liulongfang:
1. optimized uacce device query logic
2. update uadk function symbols


lin755:
lgtm

gaozhangfei:
looks good to me, and tested ok
except ./scripts/checkpatch.pl reports warning: "line over 80 characters ", though it can be ignored.
git msg
ubsequent -> subsequent

PR #509, opened 2022-10-13
doc: update release tag policy

hzhuang1:
Update the policy on release tag.

Each version followes the format of {x}.{y}.{z}. Make use of {y} to indiate the release. The release date must be mentioned in tag comments.

### DIFF
--- a/docs/maintenance.md
+++ b/docs/maintenance.md
@@ -92,10 +92,10 @@ any library with specified library version.
 The library version likes libNAME.so.{x}.{y}.{z}
 
 ```
-{x} stand for primary version, should change when APIs are changed which
+{x} stands for primary version that should change when APIs are changed which
     making things incompatible.
-{y} stand for sub version, accumulated for three monthes.
-{z} stand for mirror version, accumulated for each work week.
+{y} stands for sub version that is accumulated or reseted for each release.
+{z} stands for mirror version that is accumulated for each work week.
 ```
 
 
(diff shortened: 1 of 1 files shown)


hzhuang1:
It's the proposal on release tag policy. Any comment is welcome!

Liulongfang:
Is the version number related to the branch?

hzhuang1:
Excuse me that I didn’t get your point.

The version number in master branch (development branch) should always be libNAME.so.{x}.{y}.{z}. Only {z} is always accumulated.

For the release tag, it should be libNAME.so.{x}.{y}.0. For each release, {y} should be updated either to {y+1} or 0.

If there’s any fix on the release, it should be libNAME.so.{x}.{y}.0-[a-z].

On 14 Oct 2022, at 14:19, Liulongfang ***@***.******@***.***>> wrote:



Is the version number related to the branch?

Liulongfang:
Because the previous discussion said that it may be necessary to create a new release branch version, then this version number will be different on different branches. In other words, how to update the respective version numbers on different branches?

hzhuang1:
At first, I agreed to create a new release branch. But I concerned it may cause too much branches. So I send out my opinions on release tags.

And the most important thing is hard to create link between new date version and development version number. So I suggested to keep using development version number, but changed the meaning of {y} in version number. We could just keep the release date in tag comment.

Both of these are just my personal opinion. If it can’t resolve the problem, it could be discarded. It’s open for discussion.

On 14 Oct 2022, at 17:08, Liulongfang ***@***.******@***.***>> wrote:



Because the previous discussion said that it may be necessary to create a new release branch version, then this version number will be different on different branches. In other words, how to update the respective version numbers on different branches?

PR #510, opened 2022-10-14
doc: remove build information in README

hzhuang1:
The detailed building instructions are listed in INTALL. And the simpler one is in README.

While user reads README, he may ignore the full version of building instructions in INSTALL. So remove the simpler one.

### DIFF
--- a/README
+++ b/README
@@ -31,25 +31,6 @@ include/*
 docs/*
 	Documentations.
 
-Build UADK in native environment
-
-    $ ./cleanup.sh
-
-    Make sure that all generated files could be removed.
-
-    $ ./autogen.sh
-    $ ./conf.sh
-
-    UADK could be configured as either static or dynamic library by conf.sh.
-    By default, it's configured as dynamic library.
-
-    $ make
-    $ sudo make install
-
-    Both dynamic and static libraries would be installed in /usr/local/lib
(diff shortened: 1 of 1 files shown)


gaozhangfei:
1. do we need mention where downloading gcc
Toolchain: https://releases.linaro.org/components/toolchain/binaries/latest-7/aarch64-linux-gnu/

2. Now x86 cross-compile only numa are must required for building, do we need mention this.

Liulongfang:
If the simple compilation method in the readme is deleted, the INSTALL should be compiled for the arm platform and x86 platform in chapters, and the dependency on openssl should also be updated.

hzhuang1:
The quick build instructions in README doesn’t mention any informations on x86 platform or arm platform. So there could be a new pull request covering the suggested topics.

By the way, I think that UADK should support x86 native build. I think your meaning is on cross-platform build, not x86 native build.

On 14 Oct 2022, at 14:24, Liulongfang ***@***.******@***.***>> wrote:



If the simple compilation method in the readme is deleted, the INSTALL should be compiled for the arm platform and x86 platform in chapters, and the dependency on openssl should also be updated.

Liulongfang:
OK, cross-compilation on the new x86 platform is supported through a new pull request

PR #512, opened 2022-10-17
v1: fix x86 local build

gaozhangfei:
On x86
./cleanup.sh
./autogen.sh
./configure.sh
./v1/wd_util.h:354:2: error: #error "no platform mb, define one before

With this patch, x86 build can be succeed


Liulongfang (commented):
OK，It same to wd.h

PR #516, opened 2022-10-19
uadk: configure.ac openssl version

gaozhangfei:
UADK only works with openssl 1.1.1x, so add check


gaozhangfei:
ready for merge, verified openssl3.0  (ignored) , openssl1.0.0 (ignored) and openssl1.1.1f

PR #517, opened 2022-10-21
uadk: Replace -lcrypto with $(libcrypto_LIBS)

gaozhangfei:
Only -lcrypto may fail since no -L

Instead, we can use
$(libcrypto_LIBS): -L/usr/local/lib -lcrypto
$(libcrypto_MOCKA_CFLAGS): -I/usr/local/include

configure.ac
PKG_CHECK_MODULES(libcrypto)


gaozhangfei:
fixed commit msg, ready for merge

PR #523, opened 2022-11-01
uadk: remove with_openssl_dir

gaozhangfei:
Use standard openssl library and header,
so remove with_openssl_dir,
Replace libcrypto_wd.so with $(libcrypto_LIBS)
Replace WITH_OPENSSL_DIR with HAVE_CRYPTO


v1/test/hisi_hpre_test/Makefile.am
```diff
@@ -1,23 +1,23 @@
...
-test_hisi_hpre_times_LDADD=../../../.libs/libwd.so
+test_hisi_hpre_v1_LDADD=../../../.libs/libwd.so
+hpre_test_tools_v1_LDADD=../../../.libs/libwd.so
+test_hisi_hpre_times_v1_LDADD=../../../.libs/libwd.so
 endif
 
-test_hisi_hpre_LDADD+=$(with_openssl_dir)/libcrypto_wd.so
-hpre_test_tools_LDADD+=$(with_openssl_dir)/libcrypto_wd.so
-test_hisi_hpre_times_LDADD+=$(with_openssl_dir)/libcrypto_wd.so
+test_hisi_hpre_v1_LDADD+=$(libcrypto_LIBS)
```
hzhuang1, line 20: The original name is "libcrypto_wd.so". When you use libcrypto_LIBS, do you change the name to "libcrypto.so"?

gaozhangfei: Yes, all libcrypto_wd.so changed to libcrypto.so
libcrypto_wd.so is private and can not be built, it is the patch's purpose

hzhuang1: I noticed that Wenkai Lin changed the name from libcrytpo.so to libcrypto_wd.so. It's better to notify him first. With his  Ack, I could go on.

lin755: lgtm, the utility of this library is limited and it is not readily available, so it need to be removed. I think we may  disable test directory and add some uadk tool to test_sanity.sh in subsequent patches.

gaozhangfei:
Only change the git msg to make it clear

PR #524, opened 2022-11-01
uadk: update the symbol table for libraries

Liulongfang:
1. add dlopen error log.
2. update the libraries function symbol.


Liulongfang:
> 

OK， thanks.
2. I check and deal with it.
3. Yes, if you need to add an external API, you need to add it in the corresponding x.map

gaozhangfei:
+1

hzhuang1:
Which kind of issue do you want to fix?

Liulongfang:
This is a function required by internal security requirements iteration, protecting the external interface symbol table of the so library file

libwd.map
```diff
@@ -0,0 +1,39 @@
+UADK_2.0 {
```
hzhuang1, line 1: I prefer to follow the naming policy as "LIBWD_2.0".

https://www.gnu.org/software/gnulib/manual/html_node/LD-Version-Scripts.html

Liulongfang: We mainly reflect the UADK framework now, so the corresponding lib library interface version name is declared as UADK_2.0

PR #525, opened 2022-11-02
version: update to 2.3.38

hzhuang1:
Update version to 2.3.38


Liulongfang:
OK

PR #526, opened 2022-11-07

Liulongfang:

In addition, the interface to the APP application layer and the interface to the user mode driver layer are mixed together, making it inconvenient for the APP application layer to use UADK. After adjustment, when the APP application layer calls the uadk interface to execute services, it only needs to call the header file of the corresponding algorithm , and driver development only needs to include its own header file and the driver-side interface of the corresponding algorithm.


gaozhangfei:
build passed, but sanity_test fail

Liulongfang:
Yes! There will be a segmentation fault on this CI environment, but it test successfully on my environment, the problem on this CI I am in debug

PR #527, opened 2022-11-08

Liulongfang:

In addition, the interface to the APP application layer and the interface to the user mode driver layer are mixed together, making it inconvenient for the APP application layer to use UADK. After adjustment, when the APP application layer calls the uadk interface to execute services, it only needs to call the header file of the corresponding algorithm , and driver development only needs to include its own header file and the driver-side interface of the corresponding algorithm.

### DIFF
--- a/docs/wd_alg_init2.md
+++ b/docs/wd_alg_init2.md
@@ -1,0 +1,159 @@
+# wd_alg_init2
+
+## Preface
+
+The current uadk initialization process is:
+1.Call wd_request_ctx() to request ctxs from devices.
+2.Call wd_sched_rr_alloc() to create a sched(or some other scheduler alloc function if exits).
+3.Initialize the sched.
+4.Call wd_alg_init() with ctx_config and sched.
+
+```flow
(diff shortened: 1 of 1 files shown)


gaozhangfei:
+1

Tested OK on x86 and arm.

Two warning can be fixed, others are ok
WARNING: Missing a blank line after declarations
#715: FILE: wd_util.c:152:
+       struct wd_ctx_h *ctx = (struct wd_ctx_h *)h_ctx;
+       if (!ctx || qfrt >= UACCE_QFRT_MAX)

WARNING: suspect code indent for conditional statements (8, 24)
#715: FILE: wd_util.c:152:
+       if (!ctx || qfrt >= UACCE_QFRT_MAX)
+                       return 0;

Liulongfang:
OK! Thanks you!  I'll change it

Liulongfang:
@hzhuang1 ，Please help to review

hzhuang1:
I'm fine on this.
@haofang111 What's your opinion?

PR #528, opened 2022-11-10
uadk/v1: check dh and rsa key bits

qwl731:
Before creating ctx, the key bits must be checked. If the key bits is incorrect, return failed.

Signed-off-by: Weili Qian <qianweili@huawei.com>


Liulongfang (commented):
Looks good to me

PR #529, opened 2022-11-12

Liulongfang:

On the adjusted UADK, I divide UADK users into two categories: business users and developer users.
Business users only need to pay attention to how to use UADK to perform data acceleration services. 
Developer users only need to focus on adapting their devices to the UADK framework.

According to the different demands of these two types of users on UADK, the external interface is divided into two different headers. In this way, business users only need to obtain the required interface through “wd.h“ to fully realize their own acceleration business, and do not need to pay attention to how the device they use is driven. Developer users only need to use ”wd_util.h“ to obtain the interface required to adapt a device driver to uadk, without paying attention to how their own devices are used by users.
In this way, both types of users only need to learn and understand the interfaces in their respective header files to complete their demands.

PR reviewed before: https://github.com/Linaro/uadk/pull/527
This PR rebased on: https://github.com/Linaro/uadk/pull/520


gaozhangfei:
There is uadk_file_relationship.svg, is that required?
do we need to update this file when it is changed later?

Liulongfang:
This svg diagram is added to save the current file relationship diagram so that users can understand its structure.

At present, our file structure will not be changed after this modification. If there is an important update, it can be modified.

Liulongfang:
> There is uadk_file_relationship.svg, is that required? do we need to update this file when it is changed later?

This svg diagram is added to save the current file relationship diagram so that users can understand its structure.
At present, our file structure will not be changed after this modification. If there is an important update, it can be modified.

gaozhangfei:
> This svg diagram is added to save the current file relationship diagram so that users can understand its structure.
> At present, our file structure will not be changed after this modification. If there is an important update, it can be modified.

it will be destined to be outdated since it is too concrete.
It is meanless to update this file every time, and in fact nobody cares about this relation.
If you think it is essential, we can add the cmd on generating this pic instead.

Liulongfang:
>

Liulongfang:
> > This svg diagram is added to save the current file relationship diagram so that users can understand its structure.
> > At present, our file structure will not be changed after this modification. If there is an important update, it can be modified.
> 
> it will be destined to be outdated since it is too concrete. It is meanless to update this file every time, and in fact nobody cares about this relation. If you think it is essential, we can add the cmd on generating this pic instead.

If it is possible to generate this relationship diagram through a command, that is fine. What is this command?

gaozhangfei:
I don't know, how do you get this pic?
Some examples using cflow.

gaozhangfei:
Anyway, it's only my opinion no need to maintain this relationship pic, it is up to you.
Others are OK
+1

Liulongfang:
> I don't know, how do you get this pic? Some examples using cflow.

This pic is drawn by myself, not generated by command

Liulongfang:
> Anyway, it's only my opinion no need to maintain this relationship pic, it is up to you. Others are OK +1

My opinion is to keep it, after all, it is still beneficial for UADK users to understand UADK's software framework.
Thanks.

haofang111:
/lgtm
+1

hzhuang1:
Asked

PR #535, opened 2022-12-16
Add dynamic driver loading function

Liulongfang:
Update the current driver adaptation and usage method, from fixed
use of Hisilicon device driver to automatic registration of the
driver according to the algorithm.

When the algorithm API layer uses the driver, it is no longer
bound to the fixed device driver, but dynamically obtained
and stored according to the algorithm query to use.

Update the driver and API layer of the zip module, keep the
function of the init interface unchanged, update the implementation
of the init2 interface and match the dynamic loading function.


Liulongfang:
CI tests need to be retriggered

gaozhangfei:
git push -f can re-trigger the ci, no need to close

PR #538, opened 2022-12-28
Add dynamic driver loading function

Liulongfang:
Update the current driver adaptation and usage method, from fixed
use of Hisilicon device driver to automatic registration of the
driver according to the algorithm.

When the algorithm API layer uses the driver, it is no longer
bound to the fixed device driver, but dynamically obtained
and stored according to the algorithm query to use.

Update the driver and API layer of the zip module, keep the
function of the init interface unchanged, update the implementation
of the init2 interface and match the dynamic loading function.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -33,7 +33,7 @@ AM_CFLAGS+= -DUADK_RELEASED_TIME="\"Released ${MONTH} ${DAY}, ${YEAR}\""
 pkginclude_HEADERS = include/wd.h include/wd_cipher.h include/wd_aead.h \
 		  include/wd_comp.h include/wd_dh.h include/wd_digest.h \
 		  include/wd_rsa.h  include/uacce.h include/wd_alg_common.h \
-		  include/wd_ecc.h include/wd_sched.h
+		  include/wd_ecc.h include/wd_sched.h include/wd_alg.h
 
 nobase_pkginclude_HEADERS = v1/wd.h v1/wd_cipher.h v1/wd_aead.h v1/uacce.h v1/wd_dh.h \
 			 v1/wd_digest.h v1/wd_rsa.h v1/wd_bmm.h
@@ -41,7 +41,7 @@ nobase_pkginclude_HEADERS = v1/wd.h v1/wd_cipher.h v1/wd_aead.h v1/uacce.h v1/wd
 lib_LTLIBRARIES=libwd.la libwd_comp.la libwd_crypto.la libhisi_zip.la \
 		libhisi_hpre.la libhisi_sec.la
 
-libwd_la_SOURCES=wd.c wd_mempool.c wd.h		\
+libwd_la_SOURCES=wd.c wd_mempool.c wd.h	wd_alg.c wd_alg.h	\
 		 v1/wd.c v1/wd.h v1/wd_adapter.c v1/wd_adapter.h \
 		 v1/wd_rng.c v1/wd_rng.h	\
 		 v1/wd_rsa.c v1/wd_rsa.h	\
--- a/Makefile.am
+++ b/Makefile.am
@@ -86,7 +86,7 @@ if WD_STATIC_DRV
 AM_CFLAGS += -DWD_STATIC_DRV
 AM_CFLAGS += -DWD_NO_LOG
 
-libwd_la_LIBADD = $(libwd_la_OBJECTS) -lnuma
+libwd_la_LIBADD = $(libwd_la_OBJECTS) -ldl -lnuma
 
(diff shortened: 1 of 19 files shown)


hzhuang1:
+1

gaozhangfei:
the doc design.md load pic seems not correct.

How to use fallback, and what's the policy?
Is it transparent to the user's app (called by the drive itself) or controlled by the user?

Liulongfang:
> 
I tried to open this document and it can be opened normally. 

The fallback function can be enabled by configuring the task_type parameter of the init2 interface as TASK_MIX. Through this parameter, you can configure whether the current business is hardware computing, software instruction computing or hybrid computing

gaozhangfei:
> > 
> 
> I tried to open this document and it can be opened normally.

OK, it may my issue, I checked from your git tree.

> 
> The fallback function can be enabled by configuring the task_type parameter of the init2 interface as TASK_MIX. Through this parameter, you can configure whether the current business is hardware computing, software instruction computing or hybrid computing

Yes, driver then get two func pointers
Then how to switch them, according to packet size or cpu usage?
 is it decided from user or some policy?

Liulongfang:
> > > 
> > 
> > 
> > I tried to open this document and it can be opened normally.
> 
> OK, it may my issue, I checked from your git tree.
> 
> > The fallback function can be enabled by configuring the task_type parameter of the init2 interface as TASK_MIX. Through this parameter, you can configure whether the current business is hardware computing, software instruction computing or hybrid computing
> 
> Yes, driver then get two func pointers Then how to switch them, according to packet size or cpu usage? is it decided from user or some policy?

The fallback driver will be transparently transmitted directly to the device driver. If the current device has some algorithm or hardware limitations, which makes it impossible to handle some scenarios (such as short packets, unsupported algorithm specifications, etc.), you can forward the task to the fallback driver through the fallback driver. Therefore, how to switch the driver and when to switch the driver are all determined by the control of the device driver itself, and there is no longer an algorithm API layer for unified processing, thereby enhancing the compatibility of the device

gaozhangfei:
> > > > 
> > > 
> > > 
> > > I tried to open this document and it can be opened normally.
> > 
> > 
> > OK, it may my issue, I checked from your git tree.
> > > The fallback function can be enabled by configuring the task_type parameter of the init2 interface as TASK_MIX. Through this parameter, you can configure whether the current business is hardware computing, software instruction computing or hybrid computing
> > 
> > 
> > Yes, driver then get two func pointers Then how to switch them, according to packet size or cpu usage? is it decided from user or some policy?
> 
> The fallback driver will be transparently transmitted directly to the device driver. If the current device has some algorithm or hardware limitations, which makes it impossible to handle some scenarios (such as short packets, unsupported algorithm specifications, etc.), you can forward the task to the fallback driver through the fallback driver. Therefore, how to switch the driver and when to switch the driver are all determined by the control of the device driver itself, and there is no longer an algorithm API layer for unified processing, thereby enhancing the compatibility of the device

OK, so this is out of the user's control and decided by the driver itself.

Now the driver gets two ops, one hardware, and the other software (for example). 

Some scenarios, 
Big packets use hardware, and small packets choose software and need a threshold, which may differ according to algorithms.
Or some other scenarios like bandwidth.
Can we add it later in this framework?

Liulongfang:
>

Liulongfang:
> > > > > 
> > > > 
> > > > 
> > > > I tried to open this document and it can be opened normally.
> > > 
> > > 
> > > OK, it may my issue, I checked from your git tree.
> > > > The fallback function can be enabled by configuring the task_type parameter of the init2 interface as TASK_MIX. Through this parameter, you can configure whether the current business is hardware computing, software instruction computing or hybrid computing
> > > 
> > > 
> > > Yes, driver then get two func pointers Then how to switch them, according to packet size or cpu usage? is it decided from user or some policy?
> > 
> > 
> > The fallback driver will be transparently transmitted directly to the device driver. If the current device has some algorithm or hardware limitations, which makes it impossible to handle some scenarios (such as short packets, unsupported algorithm specifications, etc.), you can forward the task to the fallback driver through the fallback driver. Therefore, how to switch the driver and when to switch the driver are all determined by the control of the device driver itself, and there is no longer an algorithm API layer for unified processing, thereby enhancing the compatibility of the device
> 
> OK, so this is out of the user's control and decided by the driver itself.
> 
> Now the driver gets two ops, one hardware, and the other software (for example).
> 
> Some scenarios, Big packets use hardware, and small packets choose software and need a threshold, which may differ according to algorithms. Or some other scenarios like bandwidth. Can we add it later in this framework?

Yes, after the integration of this framework, there are these requirements. The HiSilicon device driver that has been enabled in UADK can add these functions, and can be identified and processed according to different algorithms

PR #543, opened 2023-01-30
version: update to 2.4.1

hzhuang1:
Update version to 2.4.1

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -21,7 +21,7 @@ endif	# WITH_LOG_FILE
 # z = revision
 MAJOR = 2
 MINOR = 4
-REVISION = 0
+REVISION = 1
 UADK_VERSION = -version-number ${MAJOR}:${MINOR}:${REVISION}
 
 DAY = 12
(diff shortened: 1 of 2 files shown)


gaozhangfei:
do we need update version?
update docs/ReleaseNotes.md as well?

Is this also required for minor release?
+DAY = 12
+MONTH = Dec
 YEAR = 2022

hzhuang1:
> do we need update version? 

Of course. We need to update the version since a new feature is submitted in.

> update docs/ReleaseNotes.md as well?
> 

Maybe Longing can handle it.

> Is this also required for minor release? +DAY = 12 +MONTH = Dec YEAR = 2022

It's just to apply a tag. I don't think it's a release.

gaozhangfei:
> > do we need update version?
> 
> Of course. We need to update the version since a new feature is submitted in.

How about waiting for the code to be stable, in case some bug fixes for this new code?


> 
> > update docs/ReleaseNotes.md as well?
> 
> Maybe Longing can handle it.

By the way, docs/ReleaseNotes.md may not be required for the minor version.
Only prepared for the next major release.


> 
> > Is this also required for minor release? +DAY = 12 +MONTH = Dec YEAR = 2022
> 
> It's just to apply a tag. I don't think it's a release.
ok

hzhuang1:
> > > do we need update version?
> > 
> > 
> > Of course. We need to update the version since a new feature is submitted in.
> 
> How about waiting for the code to be stable, in case some bug fixes for this new code?
> 
Sure. Longfang also mentioned there're a few patches for this new feature is in the coming. 
I'm waiting them.

So this tag will be applied later.

PR #544, opened 2023-02-11
uadk: update ecc/rsa/dh function

qwl731:
To simplify the calling logic of uadk, a new init2 interface is added to the ecc/rsa/dh algorithm class, which includes hardware device resource initialization and scheduler initialization.


Liulongfang (approved):
looks good to me

PR #545, opened 2023-03-01
uadk: update cipher init2 interface

Liulongfang:
In order to simplify the calling logic of uadk, a new init2 interface
is added to the cipher algorithm class, which includes hardware device
resource initialization and scheduler initialization.

### DIFF
--- a/wd_cipher.c
+++ b/wd_cipher.c
@@ -67,6 +67,17 @@ struct wd_cipher_sess {
 };
 
 struct wd_env_config wd_cipher_env_config;
+static struct wd_init_attrs wd_cipher_init_attrs;
+
+static struct wd_ctx_nums wd_cipher_ctx_num[] = {
+	{1, 1}, {}
+};
+
+static struct wd_ctx_params wd_cipher_ctx_params = {
+	.op_type_num = WD_CIPHER_DECRYPTION,
+	.ctx_set_num = wd_cipher_ctx_num,
+	.bmp = NULL,
+};
 
 #ifdef WD_STATIC_DRV
 static void wd_cipher_set_static_drv(void)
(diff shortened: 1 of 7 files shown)


wd_cipher.c
```diff
@@ -317,9 +314,92 @@ void wd_cipher_uninit(void)
...
 	wd_clear_ctx_config(&wd_cipher_setting.config);
+}
+
+int wd_cipher_init(struct wd_ctx_config *config, struct wd_sched *sched)
+{
+	bool flag;
+	int ret;
+
+	pthread_atfork(NULL, NULL, wd_cipher_clear_status);
+
```
gaozhangfei, line 347: any issue if pthread_atfork is called multi times?
if yes, how about put later, since wd_alt_try_init has protection

gaozhangfei (commented):
is init2 already in using? any example of this
        /*
         * Driver lib file path could set by env param.
         * than open tham by wd_dlopen_drv()
         * use NULL means dynamic query path
         */
        wd_cipher_setting.dlh_list = wd_dlopen_drv(NULL);

gaozhangfei:
how about sending the patch using init2 interface as well for the function test.
By the way, do we need test case for the dynamic loading module feature?

wd_cipher.c
Liulongfang: pthread_atfork is used to prevent data exceptions after multi-process fork.
The following wd_alg_try_int is used to prevent repeated initialization of multiple threads in a single process

Liulongfang:
> 

This set of init2 interface needs to re-add a set of self-test cases in the self-test script, but it is not recommended to add test code directly in this patch

gaozhangfei:
+1

PR #548, opened 2023-03-31
add adapter

gaozhangfei:
only for example

1. change drv ops api, adding para driver.
2. add func wd_alg_driver_xx to replace driver->ops.
3. drv should malloc/free ctx by itself. drv.init malloc priv ctx and drv.exit free
4. all drivers are wd_alg_driver, need remove all private driver struct like wd_aead/digest_driver 5, add wd_find_drv to find drv with drv_name and alg_name, considering several drivers supporting same alg.
ignoring refcont, to make it simple, since only used drivers in para are probed


lin755:
uadk_tool test --m sec --digest 0 --sync --optype 0 --pktlen 16 --keylen 16 --times 1 --multi 1 --init 2

PR #555, opened 2023-05-18
uadk: driver alloc and free resources by itself

gaozhangfei:
A driver should alloc and free resources by himself instead of the caller, considering multi-driver exists, the caller can not hold ctx for all drivers, so drv.init malloc private ctx and drv.exit free.

Add driver as para of alg ops, not only for getting private ctx, but easier for multi-driver support.

### DIFF
--- a/wd_cipher.c
+++ b/wd_cipher.c
@@ -59,7 +59,6 @@ struct wd_cipher_setting {
 	struct wd_sched sched;
 	struct wd_async_msg_pool pool;
 	struct wd_alg_driver *driver;
-	void *priv;
 	void *dlhandle;
 	void *dlh_list;
 } wd_cipher_setting;
@@ -318,8 +317,7 @@ static int wd_cipher_common_init(struct wd_ctx_config *config,
 		goto out_clear_sched;
 
 	ret = wd_alg_init_driver(&wd_cipher_setting.config,
-					wd_cipher_setting.driver,
-					&wd_cipher_setting.priv);
+					wd_cipher_setting.driver);
 	if (ret)
 		goto out_clear_pool;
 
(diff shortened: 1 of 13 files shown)


wd_cipher.c
```diff
@@ -323,19 +321,14 @@ static int wd_cipher_common_init(struct wd_ctx_config *config,
 
 static void wd_cipher_common_uninit(void)
 {
-	void *priv = wd_cipher_setting.priv;
-
-	if (!priv)
-		return;
-
```
Liulongfang, line 330: There is a problem when deleting directly here. After the priv is removed at the API layer, a solution needs to be added to prevent repeated initialization. Otherwise, the function of preventing double release will be gone, which will cause other problems.

Thanks.
Longfang.

Liulongfang: The priv here is still in the wd_cipher_setting structure. It is used to prevent repeated uninitialization of resources at the API layer

gaozhangfei: If the driver does not support repeated exit, it is the driver's responsibility to prevent repeated un-initialization, not the caller's.
Now both hisi_sec_init and  hisi_sec_exit has checked drv.priv.
For drivers permitted multi-init and multi-exit, it is OK, just like free().

Liulongfang: But this problem is not solved in the current patch. If merged into the patch will cause problems

gaozhangfei:
rebased to master, to solve conflict

wd_cipher.c
gaozhangfei: wd_alg_try_init(&wd_digest_setting.status) can prevent re-init2
Whether the driver can call init multi-times depends on the driver itself.

Liulongfang: Yes, this wd_alg_try_init is possible. 
You can add it and verify it.

gaozhangfei: wd_alg_try_init still there, in wd_xxx_init, this patch does not change that.

Liulongfang (approved):
looks good to me

PR #557, opened 2023-05-25
Adapter 5.25

gaozhangfei:


adapter/adapter_roundrobin.c
```diff
@@ -0,0 +1,102 @@
...
+{
+	struct uadk_adapter_ctx *ctx = (struct uadk_adapter_ctx *)adapter->priv;
+
+	if (ctx->priv) {
+		free(ctx->priv);
+		ctx->priv = NULL;
+	}
+}
+
+/* fixme, how to ensure send and recv are matched in async mode */
```
Liulongfang, line 40: There are several issues that need to be analyzed:
1. Adapter scheduling in asynchronous mode cannot currently match.
2. The current adapter’s driver scheduling can only implement the RR scheduling algorithm.
3. In a UADK framework, there are two levels of scheduling (ctx
scheduler and driver adapter) are unreasonable and should be fused

gaozhangfei: > There are several issues that need to be analyzed:
> 
> 1. Adapter scheduling in asynchronous mode cannot currently match.
Do you mean one transaction is split into several times, some use opsA, some use opsB?

> 2. The current adapter’s driver scheduling can only implement the RR scheduling algorithm.
Have realized threshold policy, but need to get packet size, via changing the struct.

> 3. In a UADK framework, there are two levels of scheduling (ctx
>    scheduler and driver adapter) are unreasonable and should be fused
They are different thing, adapter only choose ops, ie, choose driver.
The ctx is hardware resource, I think it likes memory, alloc and free, no need schedule.
Anyway, it's better to split.

Liulongfang: > > There are several issues that need to be analyzed:
> > 
> > 1. Adapter scheduling in asynchronous mode cannot currently match.
> >    Do you mean one transaction is split into several times, some use opsA, some use opsB?
>  

No. What I mean is that currently only synchronous mode is implemented in the adapter code.
In asynchronous mode, if multiple drivers are supported at the same time. It may not be possible to recycle the message.
> > 2. The current adapter’s driver scheduling can only implement the RR scheduling algorithm.
> >    Have realized threshold policy, but need to get packet size, via changing the struct.
> 

If the adapter needs to actually implement software and hardware multiple drivers to send and receive messages in a mixed manner. It needs to support the device corresponding to the driver to obtain the device load in real time. Send messages based on load data
> > 3. In a UADK framework, there are two levels of scheduling (ctx
> >    scheduler and driver adapter) are unreasonable and should be fused
> >    They are different thing, adapter only choose ops, ie, choose driver.
> >    The ctx is hardware resource, I think it likes memory, alloc and free, no need schedule.
> >    Anyway, it's better to split.
My point of view is exactly the opposite of yours. ctx is allocated from the device, and the driver is also matched one-to-one with the device. Therefore, after ctx is determined, the corresponding device is also determined. At the same time, the driver must also be determined. Therefore, when the scheduler schedules ctx, it actually completes the driver selection. Therefore, the functions of the adapter can be merged into the scheduler.

gaozhangfei: > > > There are several issues that need to be analyzed:
> > > 
> > > 1. Adapter scheduling in asynchronous mode cannot currently match.
> > >    Do you mean one transaction is split into several times, some use opsA, some use opsB?
> 
> No. What I mean is that currently only synchronous mode is implemented in the adapter code. In asynchronous mode, if multiple drivers are supported at the same time. It may not be possible to recycle the message.
> 
> > > 2. The current adapter’s driver scheduling can only implement the RR scheduling algorithm.
> > >    Have realized threshold policy, but need to get packet size, via changing the struct.
> 
> If the adapter needs to actually implement software and hardware multiple drivers to send and receive messages in a mixed manner. It needs to support the device corresponding to the driver to obtain the device load in real time. Send messages based on load data
> 
> > > 3. In a UADK framework, there are two levels of scheduling (ctx
> > >    scheduler and driver adapter) are unreasonable and should be fused
> > >    They are different thing, adapter only choose ops, ie, choose driver.
> > >    The ctx is hardware resource, I think it likes memory, alloc and free, no need schedule.
> > >    Anyway, it's better to split.
> 
> My point of view is exactly the opposite of yours. ctx is allocated from the device, and the driver is also matched one-to-one with the device. Therefore, after ctx is determined, the corresponding device is also determined. At the same time, the driver must also be determined. Therefore, when the scheduler schedules ctx, it actually completes the driver selection. Therefore, the functions of the adapter can be merged into the scheduler.

For example, the data is coming, big packet goto hardware, small packet goto sw/cpu instruction.
But we don't know the data is big or small in the beginning, so can not decide at first.

Let's discuss in the meetting with Chinese to be more clear.

PR #558, opened 2023-05-27
uadk: aead supports aes-gcm stream mode

lin755:
In stream processing encryption mode, a long file
needs to be encrypted. When the accelerator is invoked, the encryption result of each block is assembled.
The assembled result is the same as the result of
encrypting the entire file at a time.
For hisi_sec, AAD data is completed by the first BD, plaintext data are done with middle BDs and end BD. In an encrypted stream, the first BD and end BD
are unique and must be delivered to hardware.


gaozhangfei (commented):
1. how about adding test cmd in git msg, (do we need consider add test in sanity test?)
2. which func change msg->msg_state
3. if realizing soft aead drv later, is it possible to support stream mode
4. lib/crypto/aes.c is software realization, is it possible to realize software drv based on this in the future?

lin755:
> 1. how about adding test cmd in git msg, (do we need consider add test in sanity test?)
ok, I will add latter
> 2. which func change msg->msg_state
msg_state is set by user, and will not be changed
> 3. if realizing soft aead drv later, is it possible to support stream mode
I think we  need to know the algorithm very well to realize it.
> 4. lib/crypto/aes.c is software realization, is it possible to realize software drv based on this in the future?
yes，lib/crypto is designed to put all the software realization code

PR #566, opened 2023-06-08
uadk/comp - some misc fixes

youngersh:
Some misc fixes:
1. Add parameters check.
2. Fix some static check.
3. Some typo.


Liulongfang (approved):
looks good to me

PR #567, opened 2023-06-15
uadk/comp: fix duplicate release of comp

Liulongfang:
In the current version, when the resource of comp is destroyed, there is no effective blanking operation for setting's priv

therefore, this operation needs to be performed directly on the priv of the setting.


haofang111:
ok

PR #570, opened 2023-07-11
Fix two bugs for sgl

youngersh:


gaozhangfei:
when I am checking why only data_fmt == WD_SGL_BUF need free sgl and when allocating sgl

+               if (msg->req.data_fmt == WD_SGL_BUF)                                                       
+                       free_hw_sgl(h_qp, &sqe, msg->alg_type);   

I feel strange about 
                .fill_buf[WD_FLAT_BUF] = fill_buf_deflate,
                .fill_buf[WD_SGL_BUF] = fill_buf_deflate_sgl,
why need two funcs, instead of only one and check para  fmt inside the func

why not using one func fill_buf and use msg as para, in fact fill_buf_deflate already has msg as para, so in the deflate check msg->req.data_fmt could make it simpler.

static int fill_buf_deflate(handle_t h_qp, struct hisi_zip_sqe *sqe,
                            struct wd_comp_msg *msg)

Also free_hw_sgl can check return when msg->req.data_fmt == WD_SGL_BUF


Anyway this patch is OK.

youngersh:
> 
In fact, complexity is a different thing for different viewer. Maybe checking the msg->req.data_fmt to go to different branch is a simple way for coder. But in terms of the scalability and maintainability of the code, too much if-else is  hard to read and change. So here let the professionals handle the professional issues. So you can get the logic from fill_zip_comp_sqe directly. What's more, the if-else may make the funtion too long.
Back to this fix, here is a compromises for hisilicon sgl apis.  Free slg is the only one need to be done in error branch. So I add a check here.

youngersh:
> when I am checking why only data_fmt == WD_SGL_BUF need free sgl and when allocating sgl
> 
> * ```
>             if (msg->req.data_fmt == WD_SGL_BUF)                                                       
>   ```
> * ```
>                     free_hw_sgl(h_qp, &sqe, msg->alg_type);   
>   ```
> 
> I feel strange about .fill_buf[WD_FLAT_BUF] = fill_buf_deflate, .fill_buf[WD_SGL_BUF] = fill_buf_deflate_sgl, why need two funcs, instead of only one and check para fmt inside the func
> 
> why not using one func fill_buf and use msg as para, in fact fill_buf_deflate already has msg as para, so in the deflate check msg->req.data_fmt could make it simpler.
> 
> static int fill_buf_deflate(handle_t h_qp, struct hisi_zip_sqe *sqe, struct wd_comp_msg *msg)
> 
> Also free_hw_sgl can check return when msg->req.data_fmt == WD_SGL_BUF
> 
> Anyway this patch is OK.

In fact, complexity is a different thing for different viewer. Maybe checking the msg->req.data_fmt to go to different branch is a simple way for coder. But in terms of the scalability and maintainability of the code, too much if-else is hard to read and change. So here let the professionals handle the professional issues. So you can get the logic from fill_zip_comp_sqe directly. What's more, the if-else may make the funtion too long.
Back to this fix, here is a compromises for hisilicon sgl apis. Free slg is the only one need to be done in error branch. So I add a check here.

haofang111:
ok

Liulongfang (approved):
looks good to me

PR #571, opened 2023-07-31
uadk: add msg pool depth set

qwl731:
This patchset adds the message pool depth configuration. Users can input value to initialize the message  pool.


Liulongfang (commented):
looks good to me

PR #574, opened 2023-08-16
This patchset fixes some issues of the uadk.

liushangbin2233:
The first patch adds init wait timeout judgment to prevent deadlocks
when making loopback calls from the same thread.

The second and third patches add a key length validity check before
calculating the actual key length of the algorithm in xts mode. Because
if the key length is odd, it will result in incorrect calculation results.


Liulongfang (approved):
Looks good to me

PR #579, opened 2023-09-20
Upstream :fix two zip module's software issues

youngersh:
Fix two bugs.
1. Since the wd_alg_try_init change the return value, wd_zlibwrapper shoud add the return value check.
2. Support sqe_priv ops for hisilicon comp BD v3.


Liulongfang (approved):
looks good to me

PR #580, opened 2023-09-20
uadk/digest: optimize the zero byte packet checking for   hash

tq444:
Optimize the zero byte packet checking in bd2 and
 add the zero byte packet checking in bd3.
  
 Hardware v2: long hash mode does not support 0 byte packet in the
 frist and middle bd. Block hash mode does not support 0 byte packet.

 Hardware v3: long hash mode does not support 0 byte packet in the
 frist and middle bd. But, block hash mode do support 0 byte packet.


Liulongfang (approved):
looks good to me

PR #582, opened 2023-09-23
Support  AES-CBC CTS mode and user self-defined data functions

CaptainMiaow:
1. Support AES-CBC CTS(ciphertext stealing) mode in v1,
    including CBC-CS1, CBC-CS2, CBC-CS3.
2. Support user self-defined data for calculation in v1.


Liulongfang (approved):
Looks good to me

PR #583, opened 2023-09-26
support digest message carrying long data information

tq444:
uadk: fix for wd_memset_zero occupies high cpu usage
  uadk: optimize session update for long hash mode
  uadk: support digest message carrying long data information
  uadk: add out_bytes check for long hash scene
  uadk: bugfix for aead set decrypt mac for stream mode
  uadk: fix for digest mac full length check


Liulongfang (approved):
Looks good to me

PR #589, opened 2023-11-30
uadk: fix some bug

qwl731:
1.fix key transfer error issue
2.remove duplicate header files
3.fix error writing data to uninitialized memory
4.fix resource leak by add qp info clear
5.check queue status before sending doorbells


qwl731:
Re_send pull request

PR #590, opened 2023-12-01
uadk: zlibwrapper - fix return value for wd_zlib_uadk_init

youngersh:
It means the UADK compression library has been initialized when wd_comp_init2_ return -WD_EEXIST. Wd_zlibwrapper doesn't need to do it again. But WD_EEXIST is not a valid return value for zlib APIs. So change it to 0.


Liulongfang (approved):
looks good to me

PR #599, opened 2023-12-10
test: decrease the source file size from 16MB to 12MB

hzhuang1:
While decompress a compressed file, it's risky if the plain text file is 16MB. So downgrade it to 12MB.

with text file /tmp/syslog
double free or corruption (!prev)
/home/hzhuang1/test/uadk/test/sanity_test.sh: line 63: 398814 Aborted                 (core dumped) $@
Fail to run UADK tests with static build.


hzhuang1:
Close it. Use pull request #600 instead.

PR #600, opened 2023-12-12
test: disable inflate test for performance temporary

hzhuang1:
Since it's triggered a bug in hw_ifl_perf() in block mode, disable it temporarily.


gaozhangfei:
thanks Haojian, do we need to create an issue of the zip_sva_perf
+1

Liulongfang (approved):
looks good to me

PR #606, opened 2024-01-11
uadk: add secure compilation option

tq444:
Add PIE、PIC、BIND_NOW、SP、NO Rpath/RunPath、FS、Ftrapv and Strip compilation option.


Liulongfang (approved):
Looks good to me

PR #610, opened 2024-02-01
wd_rsa: fix wd_rsa_common_uninit re-entry

gaozhangfei:
Fix wd_rsa_common_uninit re-entry

Fixs: 3fc344a drivers alloc and free resources by themself

### DIFF
--- a/wd_rsa.c
+++ b/wd_rsa.c
@@ -167,6 +167,12 @@ out_clear_ctx_config:
 
 static int wd_rsa_common_uninit(void)
 {
+	enum wd_status status;
+
+	wd_alg_get_init(&wd_rsa_setting.status, &status);
+	if (status == WD_UNINIT)
+		return -WD_EINVAL;
+
 	/* uninit async request pool */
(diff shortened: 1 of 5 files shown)


wd_rsa.c
```diff
@@ -167,6 +167,12 @@ static int wd_rsa_common_init(struct wd_ctx_config *config, struct wd_sched *sch
 
 static int wd_rsa_common_uninit(void)
 {
+	enum wd_status status;
+
+	wd_alg_get_init(&wd_rsa_setting.status, &status);
+	if (status == WD_UNINIT)
+		return -WD_EINVAL;
+
```
Liulongfang, line 175: Uninit for RSA algorithm was added. Then why not add it for DH and ECC algorithms?

Liulongfang: Let’s fix the problem that priv space needs to be reallocated in this patch：https://github.com/Linaro/uadk/pull/612

Liulongfang (approved):
/lgtm

PR #612, opened 2024-02-22
uadk: drv/hisi - fix failed to init drv after fork

youngersh:
The drivers initialization function use 'drv.priv' to forbid reinit. But if the child process is forked after the parent process has initialized, it can't work due to the drivers go to wrong branch on initialization.

And the algorithms initialization function is already protected against re-entry. So it is unnecessary to check 'drv.priv' in driver.

### DIFF
--- a/drv/hisi_hpre.c
+++ b/drv/hisi_hpre.c
@@ -527,30 +527,27 @@ out:
 static int hpre_rsa_dh_init(struct wd_alg_driver *drv, void *conf)
 {
 	struct wd_ctx_config_internal *config = (struct wd_ctx_config_internal *)conf;
-	struct hisi_hpre_ctx *priv = (struct hisi_hpre_ctx *)drv->priv;
 	struct hisi_qm_priv qm_priv;
+	struct hisi_hpre_ctx *priv;
 	int ret;
-
-	if (priv) {
-		/* return if already inited */
-		return WD_SUCCESS;
-	}
 
 	if (!config->ctx_num) {
 		WD_ERR("invalid: hpre rsa/dh init config ctx num is 0!\n");
 		return -WD_EINVAL;
 	}
 
-	drv->priv = malloc(sizeof(struct hisi_hpre_ctx));
-	if (!drv->priv)
+	priv = malloc(sizeof(struct hisi_hpre_ctx));
+	if (!priv)
 		return -WD_EINVAL;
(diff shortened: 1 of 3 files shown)


gaozhangfei:
And the algorithms initialization function is already protected against re-entry
这个是指这个是吧
ret = wd_alg_try_init(&wd_cipher_setting.status);
        if (ret)
                return ret;

这个fork的时候会清0
pthread_atfork(NULL, NULL, wd_cipher_clear_status);

有道理

youngersh:
> And the algorithms initialization function is already protected against re-entry 这个是指这个是吧 ret = wd_alg_try_init(&wd_cipher_setting.status); if (ret) return ret;
> 
> 这个fork的时候会清0 pthread_atfork(NULL, NULL, wd_cipher_clear_status);
> 
> 有道理

是的，当前算法层init函数已经能防重入，驱动的init是在算法层init中调用，因为可以不用判断。就不在驱动中做同样的清理动作了。

gaozhangfei:
-       ret = hpre_init_qm_priv(config, drv->priv, &qm_priv);
+       ret = hpre_init_qm_priv(config, priv, &qm_priv);

test_hisi_hpre --trd_mode=sync --dev_path=/dev/hisi_hpre-0 会挂

gaozhangfei:
+1

drv/hisi_hpre.c
```diff
@@ -527,62 +527,56 @@ static int hpre_init_qm_priv(struct wd_ctx_config_internal *config,
...
 
-	drv->priv = malloc(sizeof(struct hisi_hpre_ctx));
-	if (!drv->priv)
+	priv = malloc(sizeof(struct hisi_hpre_ctx));
+	if (!priv)
 		return -WD_EINVAL;
 
 	qm_priv.op_type = HPRE_HW_V3_ECC_ALG_TYPE;
-	ret = hpre_init_qm_priv(config, drv->priv, &qm_priv);
+	ret = hpre_init_qm_priv(config, priv, &qm_priv);
```
Liulongfang, line 572: This problem should be caused by the patch not being fixed: https://github.com/Linaro/uadk/pull/610
After this patch is fixed, HPRE’s driver should no longer need to deal with priv issues.

Liulongfang (approved):
lgtm

PR #613, opened 2024-03-06
uadk: sample - add test case for compression

youngersh:
Please don't apply this patch


gaozhangfei:
don't know how to test uadk_comp

PR #628, opened 2024-06-17
configure: fix typo

Svelar:


gaozhangfei:
+1

by the way, where using HAVE_PERF?

gaozhangfei:
if no HAVE_PERF user, we can remove it

Liulongfang:
If it is not used, it can be deleted.
In addition, a comment needs to be added to the PR to describe what problems this patch will modify.

PR #635, opened 2024-09-05
uadk/isa-ce: fix sm4 CE driver initialization error

tq444:
CE driver of sm4 should alloc and free private ctx by himself. The cleancode issue of sm3 CE driver is also solved.


gaozhangfei:
+1

Liulongfang (approved):
looks good to me

PR #636, opened 2024-09-08
use common func wd_find_msg_in_pool to replace wd_xxx_get_msg

gaozhangfei:
use common func wd_find_msg_in_pool to replace wd_xxx_get_msg


Liulongfang:
The current PR depends on the previous PR: https://github.com/Linaro/uadk/pull/555
Now we need to discuss the branch problem of uadk first, and then we will look at it after the branch problem is determined.

gaozhangfei:
replace by 646

gaozhangfei:
reopen, since target to devel seems strange, not trigger ci

PR #638, opened 2024-10-08
[RFC] add heterogeneous computing capabilities to UADK

Liulongfang:
In the current UADK framework, the hardware acceleration function and
the software acceleration functioIn the current UADK framework, the hardware acceleration function and
the software acceleration function are merged to ensure that the software
function of instruction acceleration and the hardware function of hardware
offload can run at the same time, thus providing users with stronger performance

Under the heterogeneous scheduling mode enabled in the current scheduler,
the test performance data is as follows:

Alg		Mode(1KB)	Performance(MB/s)	CPU	
				sync		async		sync	async
sm4-ecb		init1(HW)	454		1322		100%	200.00%
		init2(HW+CE)	1445.1		1864		100%	195.00%
		increase	218.30%		41.00%		0.00%	-2.50%
sm3		init1(HW)	153.1		1481		99%	199.80%
		init2(HW+CE)	431.5		508		100%	199.80%
		increase	181.84%		-65.70%		0.91%	0.00%
------------------------------------------------------------------------

Alg		Mode(8KB)	Performance(MB/s)		CPU	
				sync		async		sync	async
sm4-ecb		init1(HW)	1407.5		9092		100%	198.00%
		init2(HW+CE)	3626.8		6021		100%	199.80%
		increase	157.68%		-33.78%		0.00%	0.91%
sm3		init1(HW)	960.4		5161.1		100%	183.80%
		init2(HW+CE)	549.6		530.1		100%	199.80%
		increase	-42.77%		-89.73%		-0.40%	8.71%
-----------------------------------------------------------------------

Without increasing the CPU usage, the performance improvement of the
synchronous mode is very huge.
In the asynchronous mode, the performance is reduced because the CPU is
used for soft calculations, which can be solved by creating dedicated
calculation threads later.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -22,12 +22,12 @@ endif	# WITH_LOG_FILE
 # y = minor
 # z = revision
 MAJOR = 2
-MINOR = 7
+MINOR = 8
 REVISION = 0
 UADK_VERSION = -version-number ${MAJOR}:${MINOR}:${REVISION}
 
-DAY = 6
-MONTH = June
+DAY = 1
+MONTH = Dec
 YEAR = 2024
 AM_CFLAGS+= -DUADK_VERSION_NUMBER="\"UADK version: ${MAJOR}.${MINOR}.${REVISION}\""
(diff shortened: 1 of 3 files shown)


Liulongfang:
Performance test results of the new framework：

PR #640, opened 2024-10-23
cipher: remove req->out_bytes req->in_bytes check

gaozhangfei:
Commit "d99d0e2 uadk: fix for cipher update iv" add check for req->out_bytes req->in_bytes, however, this breaks uadk_engine and uadk_provider,

Remove this check first, until uadk_engine and uadk_provider patches are ready.


lin755:
A patch will be sent soon to fix this.

PR #661, opened 2025-01-09
uadk: support appending tag for digest stream model

tq444:
In common digest stream mode, io_bytes and iv_bytes need to be set to 0 when the final bd is calculated. Therefore, in the appending tag scenario, need to restore the values of io_bytes and iv_bytes to the values before they are set to 0.

Therefore, the hardware can compute the overall hash value of the appending packet and the previously calculated packet, and reduce the repeated calculation.

### DIFF
--- a/v1/wd_digest.c
+++ b/v1/wd_digest.c
@@ -444,7 +444,7 @@ static int param_check(struct wcrypto_digest_ctx *d_ctx,
 			return -WD_EINVAL;
 		}
 
-		if (d_opdata[i]->has_next)
+		if (d_opdata[i]->has_next == WCRYPTO_DIGEST_DOING)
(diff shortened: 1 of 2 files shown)


tq444:
dae急需这个patch，所以master也要先合入



---Original---
From: "Zhangfei ***@***.***&gt;
Date: Thu, Jan 9, 2025 17:30 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [Linaro/uadk] uadk: support appending tag for digest stream model(PR #661)




 
这个需要么？ to master?

gaozhangfei:
ok
+1

Liulongfang (approved):
looks good to me

PR #676, opened 2025-06-19
uadk: fix a double free issue in zlibwrapper

tq444:
In the destructor, it need to check whether zlibwrapper has been initialized, otherwise it will cause double free.


Liulongfang (approved):
looks good to me

PR #685, opened 2025-08-14
Remove uadk_tool dfx --date; __DATE__ and __TIME__ cause problems

dmick:
Usage of __DATE__ and __TIME__ cause "build not reproducible" warnings on Debian (-Wdate-time) when building with dpkg-buildpackage, and since -Werror is set it breaks the build.  Just remove the functionality; it's not important enough to be worth it.

### DIFF
--- a/uadk_tool/dfx/uadk_dfx.c
+++ b/uadk_tool/dfx/uadk_dfx.c
@@ -14,7 +14,6 @@
 #include "include/wd.h"
 #include "uadk_dfx.h"
 
-#define uadk_build_date()	printf("built on: %s %s\n", __DATE__, __TIME__)
 #define ARRAY_SIZE(x)		(sizeof(x) / sizeof((x)[0]))
 #define PRIVILEGE_FLAG		0666
 
@@ -35,7 +34,6 @@ struct uadk_env_table {
 
 enum dfx_op_type {
 	DISPLAY_VERSION = 2,
-	DISPLAY_DATE,
 	DISPLAY_DIR,
 	DISPLAY_ENV,
 	DISPLAY_COUNT,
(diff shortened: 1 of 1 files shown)


Liulongfang:
My suggestion is to revise the method used to retrieve the compilation time, rather than simply removing the feature altogether.

dmick:
I don't see how the compilation time is useful to anyone, so I didn't bother to look for alternate sources.  If you have suggestions, I'm not averse.

Liulongfang:
There's a relatively simple alternative approach that can avoid using those two macros while preserving the functionality. You could try this approach to see if it works.

diff --git a/Makefile.am b/Makefile.am
index e99d62a..6e73fc5 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,9 +1,16 @@
 ACLOCAL_AMFLAGS = -I m4 -I./include
 AUTOMAKE_OPTIONS = foreign subdir-objects
+
+# get UADK build time
+BUILD_DATETIME := $(shell date -u +"%Y-%m-%d_%H:%M:%S")
+
 AM_CFLAGS=-std=gnu11 -Wall -Werror -Wextra -Wno-unused-parameter -Wfloat-equal \
          -fno-common -fno-strict-aliasing -I$(top_srcdir)/include
 AM_CFLAGS+=-fPIC -fPIE -pie -fstack-protector-strong -D_FORTIFY_SOURCE=2 \
           -O2 -ftrapv -Wl,-z,relro,-z,now -Wl,-s
+
+AM_CFLAGS += -DUADK_BUILD_DATETIME=\"$(BUILD_DATETIME)\"
+

diff --git a/uadk_tool/dfx/uadk_dfx.c b/uadk_tool/dfx/uadk_dfx.c
index 9c54b7b..8afcb15 100644
--- a/uadk_tool/dfx/uadk_dfx.c
+++ b/uadk_tool/dfx/uadk_dfx.c
@@ -14,7 +14,11 @@
 #include "include/wd.h"
 #include "uadk_dfx.h"

-#define uadk_build_date()      printf("built on: %s %s\n", __DATE__, __TIME__)
+#ifndef UADK_BUILD_DATETIME
+#define UADK_BUILD_DATETIME "unknown"
+#endif
+
+#define uadk_build_date()      printf("built on: %s\n", UADK_BUILD_DATETIME)
 #define ARRAY_SIZE(x)          (sizeof(x) / sizeof((x)[0]))
 #define PRIVILEGE_FLAG         0666

PR #759, opened 2026-04-25
uadk: use blk_nostart for zip tail flush status

yuzh1105:
When the last decompression block is fully consumed and the stream context still holds cached data, the driver asks the user to send a zero-input request to flush the remaining data from hardware.

This case is not caused by insufficient output space. The hardware treats the cached data as possibly incomplete and needs another request to continue processing, so qm_parse_zip_sqe() and qm_parse_zip_sqe_v3() should return WCRYPTO_DECOMP_BLK_NOSTART instead of WCRYPTO_DECOMP_END_NOSPACE.


IAMHCHCH:
lgtm

tomismyfriend:
lgtm

Liulongfang (approved):
looks good to me

PR #762, opened 2026-05-08
uadk: modify the return value of rsa_prepare_key() in hpre

tomismyfriend:
The 'ret' value is assigned in the case of 'req->op_type == WD_RSA_GENKEY'. When 'op_type' is not 'WD_RSA_GENKEY', 'ret' is not initialized. If 'ret' is directly returned, the caller may incorrectly determine that the operation fails. Therefore, 'WD_SUCCESS' instead of 'ret' is directly returned.


Liulongfang (approved):
looks good to me

PR #767, opened 2026-07-15
uadk/v1: modify error logging to debug level in file path check

DRBin3:
Change error logging to debug level to reduce excessive error messages in Docker environment. Now includes errno and ret values for better debugging.


Liulongfang:
looks good to me

lizhi-promax:
ok

